PR #15, opened 2023-03-10
Scoring breakdown endpoint

janlegner:


api/src/handlers/list_validators.rs
```diff
@@ -137,7 +137,7 @@ pub async fn get_validators(
         OrderField::Stake => |a: &&ValidatorRecord| a.activated_stake,
         OrderField::MndeVotes => |a: &&ValidatorRecord| a.mnde_votes.unwrap_or(0.into()),
         OrderField::Credits => |a: &&ValidatorRecord| Decimal::from(a.credits),
-        OrderField::MarinadeScore => |a: &&ValidatorRecord| Decimal::from(a.marinade_score),
+        OrderField::MarinadeScore => |a: &&ValidatorRecord| Decimal::from((a.score.unwrap_or(0.0) * 10000.0) as u64),
```
butonium, line 140: nitpick: I saw that "10000" is used in multiple places, perhaps we should set as a const and + give it a name

store/src/utils.rs
```diff
@@ -599,20 +601,133 @@ pub async fn load_validators(
     );
     update_validators_ranks(
         &mut records,
-        |a: &ValidatorEpochStats| a.marinade_score,
-        |a: &mut ValidatorEpochStats, rank: usize| a.rank_marinade_score = Some(rank),
+        |a: &ValidatorEpochStats| (a.score.unwrap_or(0.0) * 1000.0) as u64,
```
butonium, line 604: nitpick: same as 10000 comment

butonium (approved):

api/src/handlers/list_validators.rs
janlegner: Done

store/src/utils.rs
janlegner: Done

PR #18, opened 2023-03-15
Feat ⚙️ / Add workflow metrics upload endpoint

butonium:
Fix: #17

### DIFF
--- a/api/src/handlers/workflow_metrics_upload.rs
+++ b/api/src/handlers/workflow_metrics_upload.rs
@@ -37,8 +37,6 @@ pub async fn handler(
     let job_scheduled = query_params.job_scheduled.unwrap_or(false);
     let job_succeded = query_params.job_success.unwrap_or(false);
     let job_failed = query_params.job_error.unwrap_or(false);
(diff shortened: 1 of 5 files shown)


api/src/handlers/workflow_metrics_upload.rs
```diff
@@ -0,0 +1,66 @@
...
+        metrics::JOB_COUNT_SCHEDULED.inc();
+    }
+    if job_succeded {
+        metrics::JOB_COUNT_SUCCESS.inc();
+    }
+    if job_failed {
+        metrics::JOB_COUNT_ERROR.inc();
+    }
+
+    if prepare_scoring_duration != 0 {
```
janlegner, line 53: This is wrong - since you use seconds precision, 0 seems quite plausible (although very unlikely) result.

You can use more idiomatic:
```rust
if let Some(prepare_scoring_duration) = query_params.prepare_scoring_duration {
  ...
}
```
The same applies for the condition below.

janlegner (approved):

Issue #21, opened 2023-03-22
Fix ⚙️ / Change the way validator data is stored


butonium:
Due to the possibility of changing the validator identity on the run, we should store / fetch data based on validator's vote_account.

PR #22, opened 2023-03-22
Fix ⚙️ / Change the way validator data is stored

butonium:
Fix: #21 

- remove unnecessary stake caching
- replace identity with vote_account


api/src/handlers/commissions.rs
```diff
@@ -15,21 +15,21 @@ pub struct Response {
 pub struct QueryParams {}
 
 pub async fn handler(
-    identity: String,
+    vote_account: String,
```
janlegner, line 18: is this endpoint used by FE?

api/src/handlers/list_validators.rs
```diff
@@ -23,7 +27,7 @@ pub struct Response {
 pub struct QueryParams {
     epochs: Option<usize>,
     query: Option<String>,
-    query_identities: Option<String>,
```
janlegner, line 26: let's keep identities and let's have it behave as `query` if present, so we can keep FE compatible until FE switches

api/src/handlers/reports_staking.rs
```diff
@@ -38,34 +39,54 @@ fn filter_and_sort_stakes(records: &mut Vec<StakingChange>) {
...
     let cache = &context.read().await.cache;
     let mut records = Vec::new();
+    let last_epoch = match get_last_epoch(psql_client).await? {
+        Some(last_epoch) => last_epoch,
+        _ => return Ok(Default::default()),
+    };
 
     let CachedScores { scores, .. } = cache.get_validators_scores();
-    let stakes = cache.get_validators_current_stakes();
+    let validators = cache.get_validators();
```
janlegner, line 51: combine the row with the `context.read().await` - so the context lock is dropped as soon as possible

janlegner: same above for `psql_client`

api/src/handlers/reports_staking.rs
```diff
@@ -11,14 +12,14 @@ pub struct Response {
...
 #[derive(Serialize, Debug)]
 pub struct Stake {
-    identity: String,
+    vote_account: String,
     current_stake: u64,
     next_stake_sol: u64,
 }
 
 #[derive(Serialize, Debug)]
 pub struct StakingChange {
```
janlegner, line 21: let's return both fields, so FE remains compatible

janlegner:
Please also fix method `voter_max_commission_in_epoch` and prepare migration for dropping the defaults. This + backwards compatibility and then lgtm

janlegner (approved):
lfg

PR #42, opened 2023-03-31
added poor voter to unstake hints

janlegner:


store/src/scoring.rs
```diff
@@ -77,13 +77,39 @@ async fn voters_with_marinade_stake_in_epoch(
...
+        .collect())
+}
+
 pub async fn load_unstake_hints(
     psql_client: &Client,
     blacklist_path: &String,
     epoch: u64,
 ) -> anyhow::Result<Vec<UnstakeHintRecord>> {
     log::info!("Loading unstake hints in epoch: {}", epoch);
     let max_allowed_commission = 10;
```
butonium, line 111:
```suggestion
    const MAX_ALLOWED_COMMISSION: u8 = 10;
```
I think it should be a constant and placed on top of the file for better visibility

butonium (approved):

PR #67, opened 2023-06-08
Feat: improve caching

butonium:
Fix: #66


api/src/redis_context.rs
```diff
@@ -0,0 +1,18 @@
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+pub struct RedisContext {
```
janlegner, line 4: This seems very artificial, let's just use Arc RwLock redis::Client

janlegner (approved):

PR #70, opened 2023-07-04
Feat: Remove mnde collection

butonium:


janlegner (commented):
i think mnde votes should remain in the scoring endpoints - wdyt? maybe it is msol that should be there as well

butonium:
> i think mnde votes should remain in the scoring endpoints - wdyt? maybe it is msol that should be there as well

sure, then we will need to fetch msol and vemnde votes and keep them in cache and retrieve them in scoring endpoints. That decision will imply some new changes:
- adding historical data endpoint for msol/vemnde votes in snapshot-api
- add votes in cache/redis warmup process
- remove mnde votes from collector

wdyt?

PR #85, opened 2023-09-25
feat: don't show inactive validators

butonium:


api/src/handlers/list_validators.rs
```diff
@@ -135,6 +136,10 @@ pub fn filter_validators(
     mut validators: HashMap<String, ValidatorRecord>,
     config: &GetValidatorsConfig,
 ) -> Vec<ValidatorRecord> {
+    validators.retain(|_, v| {
+        v.epoch_stats.iter().take(2).any(|s| s.credits != 0 && s.activated_stake > LAMPORTS_PER_SOL)
```
janlegner, line 140: not sure about the stake condition 🤔 what about having it a > 0 at least?

also, epoch_stats may be an old record? e.g. 10 epochs old - in that case we also do not want to show the validator

janlegner (approved):

PR #86, opened 2023-09-26
Feat: Add redis tagging

butonium:


janlegner (commented):
let's make it a bit more grouped: e.g. instead of validators_tag, let's use tag_validators

e.g. ds_prod_testnet_validators, ds_prod_mainnet_validators, ds_dev_mainnet_validators

janlegner (approved):

PR #90, opened 2023-10-04
feat: add rugger info

butonium:


store/src/utils.rs
```diff
@@ -283,6 +283,89 @@ pub async fn load_versions(
     Ok(records)
 }
 
+pub async fn load_ruggers(psql_client: &Client) -> anyhow::Result<HashMap<String, RuggerRecord>> {
```
janlegner, line 286: Please add a comment describing which validators are returned by this function (e.g. which criteria is used), so we do not have read the query every time we want to know that

janlegner (approved):

PR #119, opened 2024-02-22
fix: self stake

butonium:
An improvement would be to get only the (authorized_withdrawer + delegation) slice from every stake account instead of full stake account


janlegner:
But then you would not know whether the stake is activated. It does not seem you check it here either. See https://github.com/marinade-finance/delegation-strategy-2/blob/master/collect/src/marinade_service.rs#L89

butonium:
> But then you would not know whether the stake is activated. It does not seem you check it here either. See https://github.com/marinade-finance/delegation-strategy-2/blob/master/collect/src/marinade_service.rs#L89

you're right, added that and counted only effective stake as self stake

PR #138, opened 2024-08-08
feat: add bonds as self stake

butonium:


api/src/cache.rs
```diff
@@ -119,9 +118,27 @@ pub async fn warm_validators_cache(
...
     let validators_json: String = conn.get(tagged_key).await?;
-    let validators: HashMap<String, ValidatorRecord> =
+    let mut validators: HashMap<String, ValidatorRecord> =
         serde_json::from_str(&validators_json).unwrap();
 
+    if let Ok(response) = reqwest::get(&context.read().await.bonds_url).await {
+        if response.status().is_success() {
+            if let Ok(bonds) = response.json::<BondsResponse>().await {
+                for bond in bonds.bonds {
+                    if let Some(validator) = validators.iter_mut().find(|(_, v)| v.vote_account == bond.vote_account) {
```
martinkrecek, line 128: It's a `HashMap`, why search through iterating?

store/src/dto.rs
```diff
@@ -406,6 +404,26 @@ pub struct UnstakeHintRecord {
...
+    pub pubkey: String,
+    pub vote_account: String,
+    pub authority: String,
+    pub cpmpe: Decimal,
+    pub max_stake_wanted: Decimal,
+    pub updated_at: DateTime<Utc>,
+    pub epoch: u64,
+    pub funded_amount: Decimal,
+    pub effective_amount: Decimal,
+    pub remaining_witdraw_request_amount: Decimal,
```
martinkrecek, line 423: (just a note)
Seems like we'll be stuck with `witdraw` as it spreads? 😄

api/src/cache.rs
```diff
@@ -119,9 +118,27 @@ pub async fn warm_validators_cache(
...
+                        validator.1.self_stake += bond.effective_amount;
+                    }
+                }
+            } else {
+                error!("Failed to parse bonds response JSON");
+            }
+        } else {
+            error!("Failed to fetch bonds. Status: {}", response.status());
+        }
+    } else if let Err(e) = reqwest::get(&context.read().await.bonds_url).await {
```
martinkrecek, line 138: If the request fails, we try again and do nothing if it succeeds this time, or log error if it fails again? Seems like some forgotten copy-paste or something, otherwise I don't follow what the goal would be.

martinkrecek (commented):
General question not related to specific code parts - If we fail to fetch the bonds, should we really return invalid self stakes?

I believe it would be better (for SFDP) if they got nothing/error, rather than invalid (not outdated - that would be "less wrong") numbers.

cc @janlegner (if the interface towards SFDP is driven by you)

EDIT: I checked the code a bit more, and I think a good place where to fetch the bonds, instead of here, would be when loading the validators themselves.
https://github.com/marinade-finance/delegation-strategy-2/blob/64b47e5ff5f86dbee2efba3b0f19c5fd667f1b9c/api/src/redis_cache.rs#L25-L32
That way, the validators loading would either fail or have correct data, nothing in between.

api/src/redis_cache.rs
```diff
@@ -22,13 +22,22 @@ pub async fn warm_validators(
...
         &context.read().await.psql_client,
         scoring_url,
         DEFAULT_EPOCHS,
         DEFAULT_COMPUTING_EPOCHS,
     )
     .await?;
+
+    for bond in bonds {
+        if let Some(validator) = validators.get_mut(&bond.vote_account) {
+            validator.self_stake += bond.effective_amount;
```
ochaloup, line 37: We should use the `funded_amount` for this purpose, maybe?
My understanding is that the `effective_amount` is calculated from `funded_amount` substracted by the amount that is being to be withdrawn and maybe that which is part of the settlements. Despite the amounts are not "fully active" in bonds for a particular validator stake accounts are attached to the validator and generate some revenue for him.

api/src/redis_cache.rs
```diff
@@ -22,13 +22,22 @@ pub async fn warm_validators(
...
+        }
+    }
+
     let validators_json = serde_json::to_string(&validators).unwrap();
```
ochaloup, line 41: Would not be usable to add a value `bond_effective/funded_amount` into the validator record to be part of the json and in case of some investigation we can find what the `self_stake` was originally by reducing the amount by the value taken from the bond amount value (either it is used effective or funded based of my point from prior comment).

ochaloup (commented):
All fine, just thinking if `self_stake` should not be adjusted by `funded_amount` rather than the `effective_amount`. I'm not sure how `self_stake` is used later in the pipeline flow so asking here.

butonium:
> All fine, just thinking if `self_stake` should not be adjusted by `funded_amount` rather than the `effective_amount`. I'm not sure how `self_stake` is used later in the pipeline flow so asking here.

the `self_stake` value is used to determine if the validator is passing the `100 SOL` self stake requirement

api/src/redis_cache.rs
butonium: I think this requires a rework - maybe in the future we can add this / after launch

ochaloup (approved):
:love_letter:

PR #141, opened 2024-09-05
remove redis cache and speed-up validators aggregation

janlegner:


api/src/cache.rs
```diff
@@ -229,22 +197,24 @@ pub async fn warm_cluster_stats_cache(
...
+            store::utils::load_scores(
+                &context.read().await.psql_client,
+                scoring_run.scoring_run_id,
+            )
+            .await?
+        }
+        None => Default::default(),
+    };
+    let multi_run_scores =
+        store::scoring::load_all_scores(&context.read().await.psql_client).await?;
```
martinkrecek, line 217: The fact that we have most of scores loading in `store::utils` but this is in `store::scoring` caught my attention, but considering the overall state I don't think we need to refactor this, definitely not in this PR.

martinkrecek (approved):

api/src/cache.rs
janlegner: I want to do some major re-factor of the structure and how queries are executed but agreed, not here.

migrations/0011-indexes.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX idx_validators_vote_account ON validators(vote_account);
+CREATE INDEX idx_validators_epoch ON validators(epoch DESC);
```
janlegner, line 2:
```suggestion
CREATE INDEX idx_validators_epoch ON validators(epoch);
```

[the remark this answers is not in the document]
janlegner: thanks, updated

martinkrecek (approved):

PR #142, opened 2024-09-23
feat: add epochs start/end endpoint

butonium:


api/src/handlers/epochs.rs
```diff
@@ -0,0 +1,48 @@
...
+use warp::{http::StatusCode, reply::json, Reply};
+use store::dto::EpochInfo;
+
+#[derive(Serialize, Debug, utoipa::ToSchema)]
+pub struct ResponseEpochs {
+    epochs: Vec<EpochInfo>,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct QueryParams {
```
martinkrecek, line 15: Are empty `QueryParams` needed? Not sure if it's a requirement from the framework, but it seems unnecessary/confusing.

martinkrecek (approved):
LGTM

api/src/handlers/epochs.rs
butonium: my bad I didn't close this PR after sync with LJ. This endpoint is no longer needed since the epoch's boundaries are set in stakes-etl now. Thank for review tho

PR #143, opened 2024-09-25
[bin] adding check cli for mev

ochaloup:
I was adding a CLI command to check when the last MEV collection was run and work based on that.

This is connected to further `ops-infra` change of the k8s script running MEV collect and store.

### DIFF
--- a/check/src/validators_mev.rs
+++ b/check/src/validators_mev.rs
@@ -1,27 +1,10 @@
 use log::info;
 use rust_decimal::prelude::*;
 use solana_client::rpc_client::RpcClient;
-use structopt::StructOpt;
 use tokio_postgres::Client;
 
-const MILLISECONDS_PER_SLOT: u64 = 400;
-
-#[derive(Debug, StructOpt)]
-pub struct ValidatorsMevOptions {
-    #[structopt(
-        long = "execution-interval",
-        help = "What should be the approximate time between two executions (seconds).",
-        default_value = "3600"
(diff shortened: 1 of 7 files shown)


check/src/validators_mev.rs
```diff
@@ -0,0 +1,49 @@
...
+            let epoch_data = rpc_client.get_epoch_info()?;
+            let current_epoch = Decimal::from(epoch_data.epoch);
+            let current_slot = Decimal::from(epoch_data.absolute_slot);
+
+            info!(
+                "DB stores last MEV epoch: {sql_epoch}, slot: {sql_slot}, on-chain epoch: {current_epoch}, slot: {current_slot}"
+            );
+
+            // If the current epoch is bigger than the lastly stored MEV epoch, we need to proceed
+            if current_epoch > sql_epoch {
```
janlegner, line 35: isn't this off by 1?

janlegner: (also, linguistic nitpick: `greater`)

ochaloup: Yes, you're right. Thanks. I haven't realized that while not having access to prod data. I updated the code.

ochaloup:
I'm going to merge this as it does not impact any current processing of the delegation strategy until ops-infra is updated and docker built.
I would consider the ops-infra update to be kicked into live processing just at the start of next week (not now on Friday afternoon).

PR #146, opened 2024-10-02
[store] mev store to work with sql transactions

ochaloup:
I would like to propose working with a transaction commit for storing data - either all is stored all nothing is stored.
This is a proposal for the MEV part but the same can be done for the other `store` CLI functionality.

https://github.com/marinade-finance/delegation-strategy-2/compare/with-transaction?expand=1#diff-4bed465c32fbb15b963ac515a4c3a26f2017a105d777e49c6de3954ab4167293R41

The motivation is that for the MEV cron job only part of the data was stored as the cron job was "killed" before finished.

A) WDYT?
B) Maybe there is a better way how to work with `Transaction`, `Client` and `lifetime`. I would be happy to learn.


ochaloup:
Not going to work on this. This will be solved by re-running the job and changing the checker.

PR #148, opened 2024-10-03
Fixing JITO parsing accounts

ochaloup:
On investigation with @janlegner why data is not correctly loaded into MEV delegation strategy DB table we found the collect cli does not consider the account could be either initialized with validators or loaded with data by jito (that happens at start of the "next" epoch).

This PR contains three commits:

* naming adjustment. It tries to differentiate the fact the slot that is passed from collect to store is "current slot". It's not a slot connected to epoch that the Snapshot data is loaded for.
* more logging
* fixing jito tip account parsing


butonium (approved):
lgtm

PR #150, opened 2024-10-05
disallow NULLs in mev table

janlegner:


ochaloup:
After merging the https://github.com/marinade-finance/delegation-strategy-2/pull/151
I ran the update of the schema for `mev` to permit NULLs

```
ALTER TABLE mev
ALTER COLUMN total_epoch_rewards DROP NOT NULL,
ALTER COLUMN claimed_epoch_rewards DROP NOT NULL,
ALTER COLUMN total_epoch_claimants DROP NOT NULL,
ALTER COLUMN epoch_active_claimants DROP NOT NULL;
```

This PR can be closed I believe, right @janlegner ?

ochaloup:
@janlegner nope. While I did update now (probably).

I can see now the `dev` was defined with `NOT NULL` so I updated it the same way as for prod (https://github.com/marinade-finance/delegation-strategy-2/pull/150#issuecomment-2399286420)
The `test` was defined with `IS NULLABLE YES` so I did not update it.

Are those the envs expected to be checked or I miss some other db?

Btw if the third env `test` is the correct one I should choose then I would say the schema is not following what is part of the `prod`. At least the column `epoch` is still `numeric` instead of `integer`.
But to be honest I search the `.sql`  to find the alter table on changing the column type
https://github.com/marinade-finance/delegation-strategy-2/blob/master/migrations/0000-init.sql#L35
and I cannot find that alter table in migrations. Was that omitted? @butonium 
https://github.com/marinade-finance/delegation-strategy-2/blob/master/store/src/validators_mev.rs#L113C21-L113C73

janlegner:
Thanks @ochaloup, you found correct envs, waiting for buto to respond but closing the PR as per your suggestion

PR #152, opened 2024-11-29
feat: add institutional stake [GEN-1324]

butonium:
DO NOT MERGE


collect/src/marinade_service.rs
```diff
@@ -31,6 +31,28 @@ pub fn get_marinade_stakes(
...
+        rpc_client,
+        &institutional_authority,
+        None,
+        epoch,
+        stake_history,
+    )?;
+
+    assert!(
+        !institutional_stakes.is_empty(),
+        "Failed to fetch institutional stake data"
```
ochaloup, line 51: Are we sure this is correct to expect there will be always some institutional stake?

ochaloup (changes requested):
I'm not sure about `assert` in the collect code.

Small nitpick:
I believe there is missing an item within the `api/src/handlers/config.rs` to provide info on institutional delegation authority.

ochaloup (approved):
Thanks. All good seems to me.

ochaloup:
**NOTE FOR A READER** the PR merging is postponed to be coordinated with other changes for institutional delegation strategy

PR #161, opened 2025-02-12
tls

janlegner:


.github/workflows/cargo-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+  CARGO_TERM_COLOR: always
+  # Remove unnecessary WASM build artifacts
+  WASM_BUILD_CLEAN_TARGET: 1
+  # Stripping symbols and optimizing for binary size
+  RUSTFLAGS: -C strip=symbols -C opt-level=s
+  RUSTC_WRAPPER: sccache
+  SCCACHE_GHA_ENABLED: true
+
+jobs:
+  lint-and-test:
```
martinkrecek, line 21: (non-blocking)
I'm not entirely sure (it might be run as part of some of the existing commands), but it seems to me that this only runs tests and not lint.

PR #163, opened 2025-02-17
[blacklist] enhance blacklist for mip.9

ochaloup:
Based on the MIP.9 (https://forum.marinade.finance/t/mip-9-blocklisting-malicious-validators-from-stake-auction-marketplace-to-combat-sandwich-attacks/1716) this is an addition of 73 validators that should be blacklisted from the DS SAM calculation.

I went through the code and I believe this is where data is loaded from for ds-sam and I believe that's the only place that such change needs to be changed.
https://github.com/marinade-finance/ds-sam-pipeline/blob/main/auction-config.json#L5
But please let me know if other places need to be changed.


martinkrecek (approved):
There is no other place to be updated, but to prevent any inconsistencies, please redeploy Validators API with the new image after merging.

kronael (approved):
lgtm

cf.

https://github.com/marinade-finance/ds-sam/blob/d7d0eb22cde68634293190bbcfed495b3b1b3aed/packages/ds-sam-sdk/src/sdk.ts#L86

https://github.com/marinade-finance/ds-sam/blob/d7d0eb22cde68634293190bbcfed495b3b1b3aed/packages/ds-sam-sdk/src/data-provider/data-provider.ts#L145

https://github.com/marinade-finance/ds-sam/blob/d7d0eb22cde68634293190bbcfed495b3b1b3aed/packages/ds-sam-sdk/src/data-provider/data-provider.ts#L232

PR #181, opened 2025-07-11
[priorityfee/mev] not needed to sort [GEN-5657]

ochaloup:
When I was adding sorting to the `/jito` endpoint, I thought the endpoint returned a list of data for every past epoch. But I’ve recently realised that the epoch only indicates where the data comes from — the actual purpose of the endpoint is to return the MEV rewards for the **latest available** epoch. For example, if mev vote account data is available for epoch 810, it returns that; if not, but mev vote account data is available for 809, then it returns 809 instead. The vote account will be unique in output and epoch as well.

So, sorting the data doesn't make much sense and only slows the endpoint down.
It's a "micro" optimisation, but I believe there should not be this stuff.

### DIFF
--- a/store/src/validators_jito.rs
+++ b/store/src/validators_jito.rs
@@ -466,8 +466,12 @@ pub async fn get_last_jito_info(
     psql_client: &Client,
     epochs: u64,
 ) -> anyhow::Result<Vec<JitoRecord>> {
-    let mev_records = get_last_mev_info(psql_client, epochs).await?;
-    let priority_fee_records = get_last_priority_fee_info(psql_client, epochs).await?;
+    let (mev_records, priority_fee_records) = tokio::join!(
+        get_last_mev_info(psql_client, epochs),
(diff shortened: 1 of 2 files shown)


kronael (approved):
I agree not to sort if not needed.  We should not do stuff that is not needed as a rule of thumb generally.

store/src/validators_jito.rs
```diff
@@ -466,8 +466,12 @@ pub async fn get_last_jito_info(
     psql_client: &Client,
     epochs: u64,
 ) -> anyhow::Result<Vec<JitoRecord>> {
-    let mev_records = get_last_mev_info(psql_client, epochs).await?;
-    let priority_fee_records = get_last_priority_fee_info(psql_client, epochs).await?;
+    let (mev_records, priority_fee_records) = tokio::join!(
```
kronael, line 469: Can you use try_join here?

PR #182, opened 2025-07-13
[fix rewards] considering null in /rewards

ochaloup:
This is a fix for `/rewards` endpoint falling when not ready loaded data about amounts from priority fee accounts (done by collect)

```
thread 'tokio-runtime-worker' panicked at /home/chalda/marinade/delegation-strategy-2/store/src/rewards.rs:32:21:
error retrieving column amount: error deserializing column 0: a Postgres value was `NULL`
```


ochaloup:
Closing. This should not be the fix.

PR #212, opened 2026-04-21
Create dependabot.yml [trivial]

kronael:


martinkrecek:
The pre-generated comment says to add the ignore if Rust toolchain is used, so I guess that's the case here.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: '/'
+    schedule:
+      interval: 'monthly'
+    groups:
+      actions:
+        patterns:
+          - '*'
+# NOTE: add this when using Rust toolchain in the repo
+#    ignore:
+#      - dependency-name: 'dtolnay/rust-toolchain'
```
kronael:
```suggestion
    ignore:
      - dependency-name: 'dtolnay/rust-toolchain'
```

kronael:
Like this?  Would there be any hard if its in all repos if we can't detect the ones with Rust?

martinkrecek (approved):

ochaloup:
Closes #207

PR #217, opened 2026-05-21
[validators] storing client id data (agave 4.0 update) [GEN-8006]

ochaloup:
When agave was moved to 4.0 there was added functionality to get more info about clients. The idea is placing more information into DB.

## Summary by CodeRabbit

- **New Features**
  - Added richer validator and cluster metadata, including client identity (vendor/lineage), feature set, shred version, gossip port, and RPC/pubsub availability.
  - Extended cluster stats with client diversity, client lineage, and feature-set distribution.
  - Updated API responses/DTOs and schema to expose the new stats and fields.

- **Bug Fixes**
  - Improved metadata completeness by deriving node networking and client details from a unified node inventory.
  - Made optional networking/client fields safely omit-able in serialized data.

- **Chores**
  - Added CSV parsing support for client ID mapping.


migrations/0017-cluster-nodes-fields.sql
```diff
@@ -0,0 +1,17 @@
...
+  ADD COLUMN pubsub_public BOOLEAN NULL;
+
+ALTER TABLE versions
+  ADD COLUMN client_id    TEXT     NULL,
+  ADD COLUMN client_type  TEXT     NULL,
+  ADD COLUMN feature_set  BIGINT   NULL,
+  ADD COLUMN shred_version INTEGER NULL;
+
+CREATE INDEX validators_client_type_epoch ON validators (client_type, epoch);
+CREATE INDEX validators_feature_set_epoch ON validators (feature_set, epoch);
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: We're dropping both new indexes entirely (`validators_client_type_epoch` and `validators_feature_set_epoch`).

`idx_validators_epoch` (migration `0011-indexes.sql`) already serves the `WHERE epoch = $1` filter used by `load_client_diversity_stats`/`load_feature_set_stats`, and neither query filters on `client_type`/`feature_set` (they're only grouped in application code). So the composite indexes provided no read benefit for the current access patterns.

With the indexes removed there is no index build during deployment, so the non-concurrent-creation concern is moot. Resolving.

ochaloup:
🤖 claude: Addressing two items from the review (changes applied on the branch, not yet pushed):

**`store/src/versions.rs` — persist `client_type`-only changes (outside-diff comment):** Fixed. `client_type` is now added to the `DISTINCT ON` SELECT, read back from the row, and compared in the skip predicate (`&& client_type == validator_snapshot.client_type`). A snapshot whose only change is the client classification is no longer skipped, so `versions.client_type` stays current for `load_versions`.

**`migrations/0017-cluster-nodes-fields.sql` — index column order (nitpick):** Both `validators_client_type_epoch` and `validators_feature_set_epoch` were dropped rather than reordered. `idx_validators_epoch` (migration `0011`) already covers the epoch-only filter these stats queries use, and neither query filters on `client_type`/`feature_set`, so the composite indexes had no read benefit. This also resolves the separate non-concurrent-index-creation remark.

The `JitoBam` classification comment is still under review — the suggested `bam`-before-`jito` reorder would relabel a Jito client as `agave-bam`, so we're evaluating the correct taxonomy separately.

collect/src/solana_service.rs
```diff
@@ -132,18 +128,140 @@ fn is_plausible_node_version(version: &str) -> bool {
...
+        if id.contains("bam") {
+            return ClientType::AgaveBam;
+        }
+        if id.contains("agave") || id.contains("solana") {
+            return ClientType::Agave;
+        }
+        // client_id is authoritative but this value is not one we recognise (e.g. "Unknown(10)");
+        // the raw client_id is persisted, so a future taxonomy can reclassify without a backfill.
+        return ClientType::Unknown;
+    }
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining the specific reorder — moving `bam` ahead of `jito` would classify `JitoBam` (a Jito client) as `agave-bam`, which mislabels its lineage. Per the [client-ids registry](https://github.com/solana-foundation/solana-validator-client-ids), BAM is a connection mode (`AgaveBam`, `JitoBam`, `FireBAM`), not a client family.

Instead we've changed `client_type` to track the base client lineage only and dropped the `AgaveBam` variant:

- `AgaveBam` -> `Agave`
- `JitoBam` -> `Jito`
- `FireBAM` -> `Firedancer` (the `firedancer` match was broadened to `fire`)

This also fixes a latent bug the reorder wouldn't have: `FireBAM` was previously classified as `agave-bam`. BAM participation, if needed later, can be derived from the raw `client_id` (persisted). Resolving.

ilmari-h:
checking 👀

store/src/utils.rs
```diff
@@ -1144,10 +1170,130 @@ pub async fn load_block_production_stats(
     Ok(stats)
 }
 
+pub async fn load_client_diversity_stats(
```
ilmari-h: 🤖 `load_client_diversity_stats` and `load_feature_set_stats` differ only in the grouping column name and struct field names. Consider a shared generic helper (`load_stake_diversity_by(psql, epochs, column_expr)`) returning `(stake_by, share_by, count_by, total)` to avoid drift between the two.

collect/src/solana_service.rs
```diff
@@ -132,18 +126,138 @@ fn is_plausible_node_version(version: &str) -> bool {
         })
 }
 
-pub fn get_cluster_nodes_ips(rpc_client: &RpcClient) -> anyhow::Result<HashMap<String, String>> {
-    info!("Getting cluster nodes IPs");
-    let cluster_nodes = rpc_client.get_cluster_nodes()?;
+pub fn normalize_client_type(client_id: Option<&str>, version: Option<&str>) -> ClientType {
```
ilmari-h: Would be nicer if this could use rust pattern matching instead. Guessing it's not possible because there's no strict enough format for the client id. So `contains` makes sense, just wondering.

ilmari-h (dismissed):
Lgtm, one code redundancy thing / nitpick found by Claude.

store/src/utils.rs
```diff
@@ -1204,7 +1350,8 @@ pub async fn load_validators_aggregated_flat(
                 cluster_stake AS (select epoch, sum(activated_stake) as stake from validators group by epoch),
                 cluster_skip_rate AS (select epoch, sum(skip_rate * activated_stake) / sum(activated_stake) stake_weighted_skip_rate from validators group by epoch),
                 dc AS (select validators.epoch, sum(activated_stake) / cluster_stake.stake as dc_concentration, dc_aso from validators LEFT JOIN cluster_stake ON validators.epoch = cluster_stake.epoch group by validators.epoch, dc_aso, cluster_stake.stake),
-                agg_versions AS (select vote_account, (array_agg(version order by created_at desc))[1] as last_version from versions where version is not null group by vote_account)
+                agg_versions AS (select vote_account, (array_agg(version order by created_at desc))[1] as last_version from versions where version is not null group by vote_account),
+                agg_clients AS (select vote_account, (array_agg(client_type order by created_at desc))[1] as last_client_type from versions where client_type is not null group by vote_account)
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — added `and epoch <= $2` to the `agg_clients` CTE so the latest `client_type` lookup is bounded by the requested aggregation window (change applied on the branch, pending push). The adjacent `agg_versions` CTE keeps its pre-existing unbounded "latest known" semantics; it predates this PR and was left unchanged.

store/src/utils.rs
ochaloup: 🤖 claude: Refactored as suggested (applied on the branch, pending push). Both loaders now delegate to a private helper `load_stake_distribution(psql_client, epochs, column_expr)` that owns the shared logic (last-epoch guard, per-epoch query, stake/count/share aggregation) and returns a `Vec<StakeDistribution>` with `epoch`, `total_stake`, `stake_by`, `share_by`, `count_by`. `load_client_diversity_stats` and `load_feature_set_stats` are now thin mappers into their public DTOs, so the API shapes are unchanged. One note: `column_expr` (`COALESCE(client_type, ...)` vs `COALESCE(feature_set::TEXT, ...)`) is interpolated into the SQL string since Postgres cannot bind identifiers/expressions as parameters — both call sites pass compile-time literals only. Leaving this thread open for you to review and close.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Good question — your guess is exactly right. The `client_id` reported via gossip (`getClusterNodes`) is a free-form vendor string with no enforced format, and the real-world values are composites: `JitoBam` (Jito client while connected to BAM), `FireBAM` (Firedancer + BAM), `HarmonicFrankendancer` / `HarmonicAgave` (vendor-prefixed builds), `AgavePaladin`, `Rakurai`, etc. — see the registry at https://github.com/solana-foundation/solana-validator-client-ids/blob/main/client-ids.csv. An exact `match` would need to enumerate every known literal and each new vendor variant would silently fall through to `Unknown`, whereas ordered substring checks classify by base client lineage and degrade gracefully (ordering carries the logic, e.g. `frankendancer` must be tested before `fire`). Since the raw `client_id` is persisted alongside the normalized type, a future taxonomy change can reclassify without a backfill. Leaving the thread open for you to close if this answers it.

ochaloup:
@ilmari-h Thanks for the review points. I updated the code or commented on the reasons why some points were not addressed.
When you have time, please re-review it—not urgent, though. Many thanks.

ilmari-h (dismissed):

store/src/utils.rs
```diff
@@ -1434,10 +1464,143 @@ pub async fn load_block_production_stats(
...
+        let mut stake_by: HashMap<String, u64> = Default::default();
+        let mut count_by: HashMap<String, u64> = Default::default();
+        let mut total_stake: u64 = 0;
+        for row in rows.iter() {
+            let grouping_key: String = row.get("grouping_key");
+            let activated: u64 = row.get::<_, Decimal>("activated_stake").try_into()?;
+            total_stake += activated;
+            *stake_by.entry(grouping_key.clone()).or_insert(0) += activated;
+            *count_by.entry(grouping_key).or_insert(0) += 1;
+        }
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Applied the SQL-side aggregation.

`load_stake_distribution` now runs `SELECT {column_expr} AS grouping_key, SUM(activated_stake) AS stake, COUNT(*) AS validator_count FROM validators WHERE epoch = $1 GROUP BY 1`, and the Rust loop just reads the pre-aggregated `stake`/`validator_count` per group instead of folding one row per validator.

Why: transferred rows drop from ~one per validator per epoch to ~one per grouping key, so the three groupings over `DEFAULT_CACHE_EPOCHS = 80` no longer pull ~240k rows into memory per cache warm. Output is unchanged — `total_stake` was already the sum over all rows, which equals the sum of the per-group `SUM`, and `count_by` equals `COUNT(*)` per group. `activated_stake` is `NUMERIC NOT NULL`, so `SUM` cannot be NULL for a non-empty group and the previous strictness is preserved.

Not doing the second half (one scan shared across vendor/lineage/feature-set): it changes the helper contract for a constant-factor win on a background path. Left for a follow-up if cache-warm time is ever measured as a problem.

The OpenGrep `format!()` injection flag on the same lines is a false positive here — `column_expr` only ever receives compile-time literals, which the comment above the function states.

store/src/utils.rs
ochaloup: 🤖 claude: Done — the shared helper you asked for landed as `load_stake_distribution(psql_client, epochs, column_expr)` in `store/src/utils.rs`, with a private `StakeDistribution` struct holding `epoch`, `total_stake`, `stake_by`, `share_by`, `count_by`. `load_client_diversity_stats` and `load_feature_set_stats` are now thin mappers into their public DTOs.

Why it was worth it beyond the duplication: a third grouping (`client_lineage`) was added afterwards and reused the helper unchanged, so it was a one-line addition rather than a third ~55-line clone. That is exactly the drift you were pointing at, so closing this.

collect/src/solana_service.rs
ochaloup: 🤖 claude: You were right, and the code now does what you suggested — the `contains` chain is gone.

Classification keys off the numeric gossip client id instead of the free-form string: `resolve_client_id` recovers the number (parsing `Unknown(N)`, or looking the rendered name up in a map built from the vendored `collect/client-ids.csv`), and the vendor/lineage mapping is a plain `match id { 0 => .., 1 => .., .. }`.

Why it became possible: the stable identifier turned out to be the number, not the name. The name is rendered by whichever RPC node answers `getClusterNodes`, from that node`s own compiled-in table, so the same validator can come back as `Rakurai` or as `Unknown(8)` depending on which endpoint we ask — substring matching on it was fragile for that reason, not just verbose. The only remaining string handling is stripping non-alphanumerics before lookup, because agave renders the registry names without separators (`AgaveBam` vs `Agave Bam`).

Closing this one.

ilmari-h (approved):
lgtm

martinkrecek (approved):

PR #222, opened 2026-06-08
Additional fix on non existing epoch n.2 [EVNT-37]

ochaloup:
I was reviewing the change from https://github.com/marinade-finance/delegation-strategy-2/pull/221, and I realized that I don't like the fact that if data is not available for a long time, we just skip/warn. It is ok for the data to be unavailable when stakes-etl is not done with BQ loading yet. But when it is not available for a long time, I think we should reflect that in the CLI result outcome.

With that, I'm proposing an additional change here that starts failing when we are more than 12 hours after the start of the epoch and still do not have any validator data available in BQ.

I'm not sure if it is not a little bit workaround. WDYT @martinkrecek

### DIFF
--- a/check/src/validators_jito.rs
+++ b/check/src/validators_jito.rs
@@ -4,7 +4,7 @@ use solana_client::rpc_client::RpcClient;
 use structopt::StructOpt;
 use tokio_postgres::Client;
 
-pub const MILLISECONDS_PER_SLOT: u64 = 400; // 0.4 seconds per slot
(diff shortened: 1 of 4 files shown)


martinkrecek (approved):
Makes sense to me. 👍 I think that, if anything, it's the previous change that could be called a workaround. I can't comment on the specific constant (12hrs) off the top of my head, but the approach in general looks good - warn when the data is expected to show up, fail when it already gets suspicious.

PR #228, opened 2026-07-16
[GEN-8411] extend data we return for validators

ilmari-h:
Extends the validator data the API returns for the new frontend validator views. All changes are additive (no breaking changes): new fields on existing endpoints plus one new endpoint.

New data points all from existing DB fields (PostgreSQL and BQ) with the exception of validator events. For those, added a new cronjob process that syncs the events to DB.

| Data | Endpoint | Description |
|---|---|---|
| Delegator count | `GET /validators` | `unique_delegators` — distinct delegator wallets (current epoch), sourced from BigQuery |
| Take rate | `GET /validators` | `avg_take_rate` — validator take = validator_rewards / total_rewards (inflation + MEV + block commission), 30-day window, computed directly from BigQuery at cache-warm (mirrors apy-api) |
| Downtime incidents | `GET /validators` | `incidents[]` — each DOWN interval over the last 90 epochs (`epoch`, `start_at`, `end_at`, `downtime_seconds`); frontend decides what's significant |
| Incident-free filter | `GET /validators?query_incident_free=true` | Filters to validators with zero downtime incidents in the recorded window (`incidents[]` empty); in-memory, no extra query |
| Property search | `GET /validators?query=…&search_properties=true` | With `search_properties=true`, `query` also matches datacenter location fields (`dc_country`, `dc_city`, `dc_full_city`) on top of name/vote_account/identity; in-memory, no extra query |
| Country share | `GET /validators` | `dcc_country` — validator's country share of total cluster stake |
| Location / ASN history | `GET /validators` | Per-epoch `dc_asn` / `dc_aso` / `dc_city` / `dc_country` on `epoch_stats[]` — diff consecutive epochs for an "ASN/location changed" feed (previously latest-only) |
| Take rate by epoch | `GET /validators` | Per-epoch Jito MEV + priority-fee commission (`mev_commission_bps`, `priority_commission_bps`) on `epoch_stats[]`; blend with per-epoch `commission_effective` (previously latest-only via `/jito`) |
| Country stake concentration | `GET /cluster-stats` | `dc_concentration_by_country` — stake fraction per country |
| Stake per country | `GET /cluster-stats` | `dc_stake_by_country` — total active stake per country (lamports) |
| Per-epoch events feed | `GET /validators/{vote_account}/events` (new) | Per-epoch performance (blocks produced, leader slots, skip rate, credits), downtime, and PSR settlement events (`reason`/`meta`/`amount`, raw upstream JSON) |

Ingestion is wired in ops-infra ([ops-infra#2514](https://github.com/marinade-finance/ops-infra/pull/2514)): the new `validators-events` collector runs as a scheduled cronjob and backfills history in one pass via `--from-epoch`.

## Summary by CodeRabbit

* **New Features**
  * Added a GET endpoint to list validator events by epoch, including settlements and performance context.
  * Introduced snapshot-based collection and ingestion for validator event data.
  * Enriched validator metrics with unique delegators and average take-rate using external APY data.
* **Configuration**
  * Added an `apy-api-url` option to enable take-rate enrichment.
* **Database**
  * Added a new `validators_events` table with uniqueness enforcement and performance indexes.
* **Documentation**
  * Updated the generated OpenAPI spec with the new endpoint and related response schemas.


store/src/utils.rs
```diff
@@ -526,9 +529,51 @@ pub fn update_validators_ranks<T>(
...
+
+    let mut records: HashMap<String, u64> = Default::default();
+    while let Some(row) = iter.next().await? {
+        let vote_account = row.column::<String>(0)?;
+        let unique_delegators_str = row.column::<String>(1)?;
+        records.insert(vote_account, unique_delegators_str.parse()?);
+    }
+
+    Ok(records)
+}
```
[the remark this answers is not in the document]
ilmari-h: no timeouts on BQ and apis is a convention elsewhere so leaving it.

store/src/utils.rs
```diff
@@ -526,6 +578,150 @@ pub fn update_validators_ranks<T>(
...
+                        + COALESCE(vi.amount, 0) + COALESCE(vm.amount, 0) + COALESCE(vb.amount, 0))
+                ) AS take_rate
+            FROM stakers
+            LEFT JOIN `{ds}.rewards_validators_inflation` vi
+                ON stakers.vote_account = vi.vote_account
+                AND stakers.epoch = vi.epoch AND vi.epoch >= {min_epoch}
+            LEFT JOIN `{ds}.rewards_validators_mev` vm
+                ON stakers.vote_account = vm.vote_account
+                AND stakers.epoch = vm.epoch AND vm.epoch >= {min_epoch}
+            LEFT JOIN `{ds}.rewards_validators_blocks` vb
```
honzsik: `load_take_rates` joins the three per-validator reward tables (`rewards_validators_inflation`, `rewards_validators_mev`, `rewards_validators_blocks`) directly onto the `stakers` CTE via parallel `LEFT JOIN`s on `(vote_account, epoch)` without pre-aggregating each table to one row per key first. If any of these tables has more than one row per `(vote_account, epoch)` in the 30-day window, the joins fan out and the outer `SUM()` re-counts the other tables amounts by different, epoch-varying factors, silently skewing `avg_take_rate` on `GET /validators` (no crash; `SAFE_DIVIDE` still guards div-by-zero).

Note the `stakers` CTE in this same query already pre-aggregates with `SUM(...) ... GROUP BY` — the correct pattern is applied inconsistently here.

Worth confirming against the data:
```sql
SELECT vote_account, epoch, COUNT(*) FROM rewards_validators_blocks GROUP BY 1,2 HAVING COUNT(*) > 1 LIMIT 5
```
If that returns rows, the fan-out is live.

collect/src/validators_events.rs
```diff
@@ -0,0 +1,178 @@
...
+        let vote_account = row
+            .column::<String>(1)
+            .context("Failed to parse vote_account")?;
+        let reason = row.column::<String>(2).context("Failed to parse reason")?;
+        let meta = row.column::<String>(3).context("Failed to parse meta")?;
+        let amount_str = row.column::<String>(4).context("Failed to parse amount")?;
+        let epoch: Epoch = epoch_str
+            .parse()
+            .context(format!("Failed to parse epoch '{epoch_str}' as u64"))?;
+        let amount: u64 = amount_str
```
honzsik: `amount` is parsed as `u64`, but the source and sinks are all signed: the query does `CAST(SUM(amount) AS INT64)`, the DB column is `amount NUMERIC NOT NULL` (migrations/0018-events.sql), and `SettlementRecord.amount` is a `Decimal` (store/src/dto.rs). If any `(epoch, vote_account, reason, meta)` group nets negative, `parse::<u64>()` returns `Err`, and the `?` aborts the entire `--epochs-back` collection window rather than skipping that row — which is risky for a recurring backfill job given settlement amounts keep changing over the claim window. Consider parsing as `i64`/`Decimal` to match the schema.

store/src/utils.rs
```diff
@@ -200,6 +206,52 @@ async fn get_apy_calculators(
     Ok(result)
 }
 
+/// Window (in epochs) over which per-validator downtime incidents are counted for the
+/// `incidents_count` field on `/validators`.
```
honzsik: Stale doc comment: there is no `incidents_count` field. This window feeds `load_incidents`, which populates `ValidatorRecord.incidents: Vec<IncidentRecord>` (store/src/dto.rs) — a full array of incident records, not a count. Suggest updating the comment to reference the `incidents` field so it does not mislead about the API shape.

api/src/handlers/events.rs
```diff
@@ -0,0 +1,113 @@
...
+    tag = "Validators",
+    operation_id = "List per-epoch validator events",
+    description = "Per-epoch performance, downtime and PSR settlement events for a validator. `settlements[].reason` / `.meta` are raw upstream JSON.",
+    path = "/validators/{vote_account}/events",
+    params(
+        ("vote_account" = String, Path, description = "Vote account or identity of the validator"),
+        QueryParams
+    ),
+    responses(
+        (status = 200, body = ResponseEvents)
```
honzsik: The handler returns `BAD_REQUEST` (mutually-exclusive / invalid query params), `NOT_FOUND` (unknown validator), and `INTERNAL_SERVER_ERROR` (DB/query failure), but `responses(...)` only documents `200`. Consider documenting the 400/404/500 responses. (Consistent with an earlier review note on this endpoint that appears unaddressed — though documenting only 200 does match the existing convention in other handlers.)

collect/src/validators_events.rs
```diff
@@ -0,0 +1,178 @@
...
+    let (config, _) = BqClientConfig::new_with_auth().await?;
+    Ok(BqClient::new(config).await?)
+}
+
+async fn query_validator_settlements(
+    bq_client: &BqClient,
+    table_name: &str,
+    from_epoch: Epoch,
+) -> anyhow::Result<Vec<ValidatorSettlement>> {
+    let project_table_name = format!("{GOOGLE_BQ_PROJECT_ID}.{GOOGLE_BQ_DATASET}.{table_name}");
```
janlegner, line 78: this ~is already server by validators API; + think though the caching as BQ is potentially expensive (pay per query)

ilmari-h: fixing...

store/src/utils.rs

collect/src/validators_events.rs
ilmari-h: fixed

api/src/handlers/events.rs
ilmari-h: added

collect/src/validators_events.rs
ilmari-h: yes that was expensive...

Added caching (queries only once per epoch) making the costs negligible. Confirmed we only query complete snapshots, no risk of caching stale data. Current BQ queries and budget:


| Bucket | Size × frequency | $/month |
|---|---|---|
| gate | 10 MB (floor) × every warm (4,320/mo) | $0.26 |
| delegators (MAX stakes 4.95 GB + main 0.17 GB) | ~5.12 GB × once per epoch | $0.42 |
| take-rate (main 5.17 GB + window ~14 KB) | ~5.17 GB × once per epoch | $0.42 |
| block-rewards + settlements | ~22.6 MB × once per epoch | $0.003 |
| **TOTAL** | | **≈ $1.10 / month** |

honzsik:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

honzsik (approved):

PR #231, opened 2026-07-24
[GEN-8457] [GEN-8423] Verified filtering for validators list, filter by flagged and pagination

ilmari-h:
Depends on this PR that adds the verified validators list to validator bonds api: https://github.com/marinade-finance/validator-bonds/pull/478

Exposes wether validator is verified or not via `/validators` endpoint

Also adds a fix for `/validators` endpoint where it previously returned a field `validators_aggregated` which contained aggregate stats for all validators, not just the ones returned. Modified it so that `validators_aggregated` is also affected by filters and pagination, and only reflects stats of the returned validator rows.

## Summary

* **New Features**
  * Added validator list filters for verified and flagged/unflagged results, with stable take-rate sorting.
  * Added `total_count` to list responses and expanded validator details with `verified` status from the validator-bonds service.
  * Added `--validator-bonds-api-url` / `VALIDATOR_BONDS_API_URL` configuration.
  * Expanded cluster stats with additional per-epoch diversity/lineage/feature-set breakdowns.

* **Bug Fixes**
  * List-related aggregated results now match the filtered and paginated page.

* **Tests**
  * Added unit tests for flagged filtering and deterministic sorting behavior.


store/src/utils.rs
```diff
@@ -758,9 +758,32 @@ pub async fn load_take_rates() -> anyhow::Result<HashMap<String, f64>> {
...
 
+#[derive(serde::Deserialize)]
+struct VerifiedValidatorsResponse {
+    verified_validators: Vec<String>,
+}
+
+// `base` is the validator-bonds API base URL; the verified path is appended here.
+pub async fn load_verified_validators(base: &str) -> anyhow::Result<HashSet<String>> {
+    let url = format!("{}/validators/verified", base.trim_end_matches('/'));
+    let resp = reqwest::get(&url).await?;
```
ochaloup: 🤖 claude: `reqwest::get` here has no timeout; since `spawn_cache_warmer` awaits this sequentially before looping, a hang would freeze refreshes for all cache types, not just verified status.

rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
   channel = "1.88.0"
+  components = ["rust-analyzer"]
```
ochaloup: 🤖 claude: Missing trailing newline — violates this repo's `.editorconfig` (`insert_final_newline = true`).

api/src/handlers/list_validators.rs
```diff
@@ -240,6 +256,14 @@ pub fn filter_validators(
         validators.retain(|_, v| v.incidents.is_empty() == query_incident_free);
     }
 
+    if let Some(query_verified) = config.query_verified {
+        validators.retain(|_, v| v.verified == query_verified);
```
ochaloup, line 267: 🤖 claude: New `query_flagged` filter (XOR-shaped condition) and the sort tiebreak have no test coverage to catch a future logic inversion.

ochaloup (commented):
🤖 claude: Automated review found one medium-severity availability risk (unbounded HTTP call in the cache warmer) and two low-severity nits (missing trailing newline, missing test coverage for new boolean filter logic) — see inline comments.

---

@ilmari-h 
I think this PR should be rebased onto `main` (see https://github.com/marinade-finance/delegation-strategy-2/pull/232), finalized, and merged. Once it's done, I would like to put my further optimizations on top, as they are prepared at https://github.com/marinade-finance/delegation-strategy-2/pull/233.

---

```
1. api/src/cache.rs — real semantic conflict, not just textual

<<<<<<< performance-enahnce-2
        ctx.cache.validators = validators;
        ctx.cache.validators_aggregated = validators_aggregated;
=======
        ctx.cache.validators.clone_from(&validators);
>>>>>>> gen-8423-and-gen-8457

- performance-enahnce-2 keeps caching a global validators_aggregated at warm-time (computed once over all validators) — that's exactly the bug #231 fixes.
- gen-8423-and-gen-8457 deletes the validators_aggregated field/type/getter from Cache entirely, because it now recomputes the aggregate per-request from only the filtered+paginated page (in list_validators.rs, which merges in cleanly).
- Knock-on compile break: #231 also changed aggregate_validators's signature from &HashMap<String, ValidatorRecord> to &[ValidatorRecord]. Since performance-enahnce-2 never touches that function, the merge silently adopts the new signature — but cache.rs's leftover aggregate_validators(&validators) call
still passes a &HashMap. A naive merge won't compile.
- Correct resolution: take #231's complete removal of the cache-level aggregate (field, type, getter, and the aggregate_validators call in cache.rs), but keep performance-enahnce-2's ctx.cache.validators = validators; (move instead of clone) for the line that remains.

2. api/src/handlers/list_validators.rs — convergent refactor, safe to resolve

Both branches independently rewrote get_validators/filter_validators the same way (own the HashMap via cache.get_validators(), filter_validators takes ownership + retain, instead of borrowing and cloning per row — this is performance-enahnce-2's "drop redundant clones" commit). They only diverge where #231
goes further:
- return type Vec<ValidatorRecord> → (Vec<ValidatorRecord>, usize) (adds total_count for pagination)
- adds a vote_account tiebreak in sort_by (needed so paginated pages stay stable across cache refreshes)

Since #231's version is a strict superset of performance-enahnce-2's optimization, resolving = take theirs; no perf work is lost. A second, trivial conflict a few lines down (the new query_verified/query_flagged retain filters) is a pure insertion — also just take theirs.

Bottom line

No real feature-vs-feature fight — it's one genuine correctness-vs-optimization collision (the cached aggregate) plus one coincidental double-implementation of the same refactor. Recommended order: rebase gen-8423-and-gen-8457 onto current master first (closes the 3-PR gap), then merge/rebase
performance-enahnce-2 on top, resolving cache.rs by deleting the dead global-aggregate cache and list_validators.rs by taking #231's version of get_validators.
```

ochaloup:
I'm sorry, I did not want to close; I wanted to comment (see above) :-)

store/src/utils.rs
ilmari-h: added gracious timeout window.

rust-toolchain.toml
ilmari-h: fixed

api/src/handlers/list_validators.rs
```diff
@@ -240,6 +263,14 @@ pub fn filter_validators(
...
+    }
+
+    if let Some(query_flagged) = config.query_flagged {
+        validators.retain(|_, v| v.warnings.is_empty() != query_flagged);
+    }
```
ochaloup, line 272: 🤖 claude: `query_verified` has no test coverage while `query_flagged` right below it has three, and the two filters use opposite operators (`==` vs `!=`). Both are correct today, but normalising them to a single operator later would silently invert `?query_verified=true` into returning only unverified validators, with `fmt`/`clippy`/all tests still green. The `validator()` test helper hardcodes `verified: false`, so it needs a `verified` parameter before the two mirroring tests can be added.

api/src/handlers/list_validators.rs
```diff
@@ -90,19 +97,13 @@ pub struct GetValidatorsConfig {
 pub async fn get_validators(
     context: WrappedContext,
     config: GetValidatorsConfig,
-) -> anyhow::Result<Vec<ValidatorRecord>> {
+) -> anyhow::Result<(Vec<ValidatorRecord>, usize)> {
     let validators = context.read().await.cache.get_validators();
 
     let mut validators = filter_validators(validators, &config);
+    let total_count = validators.len();
```
ochaloup, line 104: 🤖 claude: `total_count` is correctly taken after filtering and before `skip`/`take`, but no test pins that ordering. Moving it below the pagination would make it equal the page size for every request, silently capping a paginating frontend at a single page.

api/src/cache.rs
```diff
@@ -187,6 +181,7 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
     let validators = store::utils::load_validators(
         &context.read().await.psql_client,
         context.read().await.scoring_url.clone(),
+        context.read().await.validator_bonds_api_url.clone(),
```
ochaloup, line 184: I have got a response from @janlegner here https://marinade-group.slack.com/archives/C0BJZAZC66S/p1785159279986539
but if it is important to load the verified validators here, why are we managing the YAML list under the Validator Bonds API and coupling this service directly to bonds?

ochaloup: btw. @ilmari-h do you plan update in `ops-infra` to point dev/prod instances of bonds API URL?

ochaloup (approved):
It seems all fine to me. Just asking about validators bonds API dependency that is coupling this service to another service - is it needed (a good way to do)?

Another question is for `GEN-8423` that talks about "MEV-inclusive net APY" and that's not part of the code here in PR, isn't it?

api/src/cache.rs
ilmari-h: yes will update it

ilmari-h:
> Another question is for `GEN-8423` that talks about "MEV-inclusive net APY" and that's not part of the code here in PR, isn't it?

we're doing that in apy-api: https://github.com/marinade-finance/apy-api/pull/65

api/src/handlers/list_validators.rs
ilmari-h: fixed

api/src/handlers/list_validators.rs
ilmari-h: nit

PR #232, opened 2026-07-28
[perf] lower number of queries [GEN-8006]

ochaloup:
A follow-up to https://github.com/marinade-finance/delegation-strategy-2/pull/217

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced database round-trips for stake distribution loading across epoch ranges.
  * Streamlined validator aggregation and client/version metadata calculations.

* **Bug Fixes**
  * Improved correctness of client metadata and version selection at epoch boundaries.
  * Ensured zero-epoch requests return no results and gaps are handled with zero/empty outputs.

* **Tests**
  * Added PostgreSQL-backed integration tests that validate stake distribution, diversity, and concentration stats.
  * Updated CI to provision PostgreSQL for the test suite and support local connections.

* **Documentation**
  * Updated development guidance with steps and `DS_TEST_POSTGRES_URL` for running SQL loader tests.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -29,3 +29,19 @@ docker cp postgresql-${DB}:/etc/ssl/certs/ssl-cert-snakeoil.pem /tmp/postgres-ro
 ```
 
 The PostgreSQL URL is then `postgresql://delegation-strategy:delegation-strategy@localhost:5432/delegation-strategy`
+
+# 2. Run the SQL loader tests
+
+`store/tests/cluster_stats_sql.rs` exercises the `/cluster-stats` and
+`/validators/flat` queries against a real PostgreSQL. It creates its own schema,
+applies `migrations/*.sql` into it and drops it again, so it needs an empty
+database only on the first run.
+
+```bash
+export DS_TEST_POSTGRES_URL='postgresql://delegation-strategy:delegation-strategy@localhost:5432/delegation-strategy'
+cargo test --all-features
+```
+
+Without `DS_TEST_POSTGRES_URL` the tests report why they are skipped and pass.
+On Podman, start the container with `podman run` and export
+`DOCKER_HOST=unix:///run/user/1000/podman/podman.sock`.
--- a/store/src/utils.rs
+++ b/store/src/utils.rs
@@ -1484,27 +1484,38 @@ async fn load_stake_distribution(
     };
     let first_epoch = last_epoch - epochs.min(last_epoch) + 1;
 
-    let mut distributions: Vec<StakeDistribution> = Default::default();
-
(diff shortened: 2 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -29,3 +29,19 @@ docker cp postgresql-${DB}:/etc/ssl/certs/ssl-cert-snakeoil.pem /tmp/postgres-ro
 ```
 
 The PostgreSQL URL is then `postgresql://delegation-strategy:delegation-strategy@localhost:5432/delegation-strategy`
+
+# 2. Run the SQL loader tests
+
+`store/tests/cluster_stats_sql.rs` exercises the `/cluster-stats` and
+`/validators/flat` queries against a real PostgreSQL. It creates its own schema,
+applies `migrations/*.sql` into it and drops it again, so it needs an empty
+database only on the first run.
```
martinkrecek, line 38: (non-blocking - minor)
The code does `DROP SCHEMA IF EXISTS` first, so I'd say it's not needed even on the first run (schema permissions are enough).

store/src/utils.rs
```diff
@@ -1656,14 +1667,16 @@ pub async fn load_validators_aggregated_flat(
...
-                agg_versions AS (select vote_account, (array_agg(version order by created_at desc))[1] as last_version from versions where version is not null group by vote_account),
-                agg_clients AS (select vote_account, (array_agg(client_vendor order by created_at desc) filter (where client_vendor is not null))[1] as last_client_vendor, (array_agg(client_lineage order by created_at desc) filter (where client_lineage is not null))[1] as last_client_lineage from versions where epoch <= $2 group by vote_account)
+                agg_versions AS (select vote_account, (array_agg(version order by created_at desc) filter (where version is not null))[1] as last_version, (array_agg(client_vendor order by created_at desc) filter (where client_vendor is not null and epoch <= $2))[1] as last_client_vendor, (array_agg(client_lineage order by created_at desc) filter (where client_lineage is not null and epoch <= $2))[1] as last_client_lineage from versions group by vote_account)
```
martinkrecek, line 1679: (non-blocking - further optimization)
Could also `cluster_stake`, `cluster_skip_rate` and `dc` be "inlined" here? They get filtered to the epoch range in the end too, so adding the epoch filter here for them should speed it up further, I think.

martinkrecek (approved):

store/src/utils.rs
ochaloup: Will be fixed in a follow-up PR

DEVELOPMENT.md
ochaloup: Will be fixed in a follow-up PR

PR #233, opened 2026-07-29
Performance enhancements [GEN-8006]

ochaloup:
While working on adding delegation client IDs and related enhancements, several more pain points were observed. This PR tries to improve the behaviour further.

Related:

* https://github.com/marinade-finance/delegation-strategy-2/pull/217
* https://github.com/marinade-finance/delegation-strategy-2/pull/232

- `/validators` and the four per-validator endpoints no longer deep-clone the entire validator cache on every request.
- Each API replica gets its own cache warm-up phase instead of all replicas warming on the same wall-clock boundary.
- Adds `migrations/0019-vote-account-indexes.sql` so the per-minute collector queries stop doing a full scan plus full sort of the whole table.
- Removes two duplicate queries per warm cycle.
- Stops the cache warmers deep-copying freshly loaded data when they could move it, including one copy made while holding the context write lock.
- Rank assignment no longer re-scans each validator's epoch list to find the entry it just ranked.
- Bounds the `cluster_stake` / `cluster_skip_rate` / `dc` CTEs to the requested epoch window so `/validators/flat` stops aggregating all of history to serve ten epochs (raised in review on #232).
- `/cluster-stats` block production now covers the same epoch window as the other four series — **this changes the response shape**, adding the oldest requested epoch and zero-filling epochs with no validator rows.
- Adds test coverage for the paths this branch touches that had none: validator filtering and paging, cache key lookup, warm-up scheduling, rank assignment, and the epoch-window bounds.
- Corrects the `DEVELOPMENT.md` claim that the SQL tests need an empty database, since they drop and recreate their own schema every run (raised in review on #232).

## Deploying

`0019` should be pre-built with `CREATE INDEX CONCURRENTLY` before the migration is applied, because `versions` (4M rows) and `uptimes` (3.2M rows) have writers every minute; the migration uses `IF NOT EXISTS` so it becomes a no-op afterwards.

## Summary by CodeRabbit

* **Performance**
  * Improved validator lookup speed across commissions, events, uptimes, and versions.
  * Added database indexes to accelerate validator and time-based queries.
  * Improved cache warming reliability and scheduling across hosts.

* **Bug Fixes**
  * Corrected validator ranking across epochs and tied rankings.
  * Ensured missing epochs display zeroed block-production metrics.
  * Restricted aggregated statistics to the requested epoch range.

* **Documentation**
  * Clarified that SQL loader tests safely recreate their schemas and can run repeatedly.

### DIFF
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -394,6 +394,134 @@ mod tests {
         );
     }
 
+    // Context owns a live Client, so get_validators cannot be exercised without a database
+    // even though it never queries one
+    async fn context_with(records: Vec<ValidatorRecord>) -> Option<WrappedContext> {
+        let url = std::env::var("DS_TEST_POSTGRES_URL").ok()?;
+        let (client, connection) = tokio_postgres::connect(&url, tokio_postgres::NoTls)
+            .await
+            .unwrap();
+        tokio::spawn(async move {
+            let _ = connection.await;
+        });
+
+        let mut context = crate::context::Context::new(
+            client,
+            Default::default(),
+            Default::default(),
+            Default::default(),
+        )
+        .unwrap();
+        context.cache.validators = cache_of(records);
+        Some(std::sync::Arc::new(tokio::sync::RwLock::new(context)))
+    }
+
+    fn stake_record(vote_account: &str, epochs: &[u64], stake: u64) -> ValidatorRecord {
+        let mut record = record(vote_account, &format!("id-{vote_account}"), epochs);
+        record.activated_stake = Decimal::from(stake);
+        record
+    }
+
+    #[tokio::test]
+    async fn get_validators_orders_by_stake_and_pages_without_touching_the_rest() {
+        let records = vec![
+            stake_record("small", &[LAST_EPOCH - 1, LAST_EPOCH], 10),
+            stake_record("large", &[LAST_EPOCH - 1, LAST_EPOCH], 30),
+            stake_record("medium", &[LAST_EPOCH - 1, LAST_EPOCH], 20),
+        ];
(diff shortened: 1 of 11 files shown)


ochaloup:
...waiting for https://github.com/marinade-finance/delegation-strategy-2/pull/231 being merged

api/src/handlers/list_validators.rs
```diff
@@ -91,18 +91,16 @@ pub async fn get_validators(
     context: WrappedContext,
     config: GetValidatorsConfig,
 ) -> anyhow::Result<Vec<ValidatorRecord>> {
-    let validators = context.read().await.cache.get_validators();
+    let ctx = context.read().await;
```
ilmari-h, line 94: If I understand correctly, does this now cause the lock to be held for the duration of the request? 

Whereas previously, it was dropped after this first line:

```
let validators = context.read().await.cache.get_validators();
```

is it ok for it to hold the lock for the duration of the whole function?

api/src/handlers/list_validators.rs
```diff
@@ -154,10 +146,10 @@ fn get_field_extractor(order_field: OrderField) -> Box<dyn Fn(&ValidatorRecord)
     }
 }
 
-pub fn filter_validators(
-    mut validators: HashMap<String, ValidatorRecord>,
+pub fn filter_validators<'a>(
+    validators: &'a HashMap<String, ValidatorRecord>,
     config: &GetValidatorsConfig,
-) -> Vec<ValidatorRecord> {
+) -> Vec<&'a ValidatorRecord> {
```
ilmari-h, line 152: Nice, less cloning!

ilmari-h (approved):
1 question but looks good!

PR #237, opened 2026-08-03
[GEN-8626][api] adding readiness probe for cleaner deployment

ochaloup:
Replaces ds-api's readiness heuristic with real probes, so a pod joins the Service only once every cache it serves has loaded.

- Add `/readyz` backed by a latched flag held outside the context lock, so all six warm steps must succeed before a pod is Ready
- Add a dependency-free `/healthz` liveness check, since ds-api had none and probing the DB would kill pods for a slow upstream
- Retry only the warm steps that failed, with capped backoff and only while cold, so a failing upstream cannot amplify load
- Exit the process if the cache warmer stops, since both probes otherwise stay green on a frozen cache
- Keep a populated cache when a refresh returns no validators, while still letting a freshly bootstrapped environment reach ready
- Record `ds_cache_last_success_seconds` per cache so a stalled refresh stays alertable without flipping replicas to 503

See https://github.com/marinade-finance/ops-infra/pull/2624

## Summary by CodeRabbit

* **New Features**
  * Added `/healthz` endpoint for liveness checks, returning `200 OK` with `alive`.
  * Added `/readyz` endpoint to report API readiness.
  * Returns `200 OK` when required data is loaded, or `503 Service Unavailable` while warming up.
  * Added automatic cache-loading retries, scheduled refreshes, and cache success timestamps for monitoring.

* **Bug Fixes**
  * Preserved existing valid cache data when a refresh returns no validators.
  * Prevented readiness from being reported before cache loading completes.


api/src/cache.rs
```diff
@@ -333,39 +365,143 @@ pub async fn warm_scores_cache(context: &WrappedContext) -> anyhow::Result<()> {
...
+                pending[index] = false;
+                record_success(name);
             }
+            Err(err) => error!("Failed to update the {name}: {err}"),
+        }
+    }
+}
+
+pub fn spawn_cache_warmer(context: WrappedContext, ready: ReadyFlag) {
+    tokio::spawn(async move {
```
ilmari-h, line 583: ## 🤖 A dead warmer task leaves both probes green forever

`tokio::spawn` swallows panics. If this task dies, `ready` stays latched
(`cache.rs:62` — nothing clears it) so `/readyz` keeps returning 200, and
`/healthz` returns "alive" unconditionally (`handlers/health.rs:13`). Pod serves a
frozen cache forever, never restarted, never pulled from the Service.

Panic paths exist today: `cache.rs:391` and `:396` both `.unwrap()` on
`duration_since(UNIX_EPOCH)`.

Fix: crash the process, same as the metrics server already does at `metrics.rs:97`
— wrap the loop in an inner `tokio::spawn`, `.await` the `JoinHandle`, and
`std::process::exit(1)` on error. Or make `/readyz` read the last-success
timestamp instead of a latch, which also covers a warmer wedged on a hung DB.

Either way the `ds_cache_last_success_seconds` alert needs to ship too, and must
use `absent()` — `prometheus::gather()` won't export the gauge until the first
success.

api/src/cache.rs
```diff
@@ -172,22 +193,33 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
...
+            DEFAULT_CACHE_EPOCHS,
+            DEFAULT_COMPUTING_EPOCHS,
+            &unique_delegators,
+            &take_rates,
+        )
+        .await?
+    };
+
+    // Bail before the write so an empty result cannot overwrite a populated cache.
+    anyhow::ensure!(
```
ilmari-h: ## 🤖 empty-validators `ensure!` bricks new environments

Any DB with zero validators (fresh staging, new cluster, local bootstrap) fails
this step every run, so `pending[5]` never clears, `mark_ready` (`:431`) is never
reached, and `/readyz` returns 503 forever. Rollout hangs instead of failing
informatively — and only on the environments where that's hardest to debug.

Anti-clobber is the right goal, but it should guard the write, not the step
result:

```rust
    if validators.is_empty() {
        let populated = !context.read().await.cache.validators.is_empty();
        anyhow::ensure!(!populated, "no validators loaded, keeping previous cache intact");
        info!("No validators in DB, nothing to cache");
        return Ok(());
    }
```

Warm cache still errors and retries; empty environment can reach ready.

If "never serve empty validators" is intentional, fine — but then `:319`
(`warm_scores_cache` returns `Ok` with `Default::default()` when there's no
scoring run) needs the same treatment, or the pod latches ready serving empty
`/validators/scores` anyway.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — the warmer loop now runs in an inner `tokio::spawn` whose `JoinHandle` is awaited, so a panic logs and `std::process::exit(1)`s exactly like the metrics server already does. The `absent()`-based `ds_cache_last_success_seconds` alert is tracked separately rather than here, since neither this repo nor ops-infra holds any alert rules as code today.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed by guarding the write rather than the step result — the `ensure!` now fires only when the in-memory cache is already populated, so a warm pod still refuses to clobber real data while an empty DB caches an empty set and reaches ready (verified against a fresh local DB: `/readyz` 200, `/validators` `total_count` 0). That also settles the `warm_scores_cache` half of your comment: empty scores returning `Ok` is now consistent with empty validators, so no change is needed there.

ochaloup:
@ilmari-h thank you for the review. I updated the code if you may re-review :pray:

ochaloup:
🤖 claude: responses to the CodeRabbit review-body findings, which have no inline thread to resolve.

**Full retry re-runs every loader (`cache.rs` 369-431)** — already fixed: a named step table plus a pending array means only the steps that failed are retried, with backoff doubling to a cap.

**Response body does not surface which cache is degraded (`readiness.rs` 14-22)** — declined, as the suggestion pre-dates the current design: `CacheLoaded` no longer exists and the handler holds only an `AtomicBool`, so exposing per-category state would put shared state back behind the context lock purely to duplicate `ds_cache_last_success_seconds`.

**Warm independent steps concurrently (`cache.rs` 402-415)** — declined: `futures` is indeed already a dependency, but `Context` holds a single `psql_client`, so the queries serialise on that one connection regardless and concurrency would only widen guard overlap.

**Ticket reference in PR title** — the check quoted an older title; `[GEN-8626]` is present, just not in the leading position.

api/src/cache.rs
```diff
@@ -483,39 +515,88 @@ pub async fn warm_scores_cache(context: &WrappedContext) -> anyhow::Result<()> {
...
+                sleep(Duration::from_secs(seconds_until_next_window())).await;
             }
+        });
 
-            let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
-            let run_every = Duration::from_secs(CACHE_WARMUP_TIME_S);
-            let sleep_seconds = now.as_secs() % run_every.as_secs();
-            sleep(Duration::from_secs(run_every.as_secs() - sleep_seconds)).await;
+        // Neither probe can observe a stopped warmer, so only process death sheds the frozen cache.
+        if let Err(err) = warmer.await {
```
ilmari-h, line 613: ### 🤖 Warmer crash-on-death only covers panics
`warmer.await` resolves on panic/abort only. A step hung on a query or a timeout-less HTTP call keeps the task alive: liveness 200, ready latched, cache frozen forever.
Fix: `tokio::time::timeout` per step, or a last-success watchdog.

api/src/cache.rs
```diff
@@ -338,6 +360,16 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
     )
     .await?;
 
+    // A freshly bootstrapped environment has no validators yet and must still be able to reach ready.
+    if validators.is_empty() {
```
ilmari-h, line 366: ### 🤖 Cold pod can go ready with an empty cache while the DB is full
`load_validators` returns `Ok(empty)` whenever `MAX(epoch) FROM cluster_info` is NULL (`store/src/utils.rs:877`), not only on a fresh env. Empty `cluster_info` + full `validators` table = pod latches ready serving nothing. The `cache.validators.is_empty()` guard can't catch it: a cold cache is empty by definition.
Fix: make it explicit — env flag, or `SELECT EXISTS(SELECT 1 FROM validators)`.

api/src/metrics.rs
```diff
@@ -75,6 +75,13 @@ lazy_static! {
     pub static ref JOB_DURATION: IntGaugeVec =
         register_int_gauge_vec!("ds_job_duration", "Workflow jobs duration", &["workflow"])
             .unwrap();
+    // Readiness latches once warm, so a cache that stops refreshing is only visible here.
```
ilmari-h, line 78: ### 🤖 `ds_cache_last_success_seconds` missing until first success
`lazy_static` registers on first deref in `record_success`, so a pod that never succeeds exports no series and the staleness alert never fires. Initialize all six labels at startup.

api/src/cache.rs
```diff
@@ -338,6 +360,16 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
...
+        let populated = !context.read().await.cache.validators.is_empty();
+        anyhow::ensure!(
```
ilmari-h, line 368: ### 🤖 Bail discards good upstream data
`ensure!` fires after bonds/APY/BigQuery already loaded, before the write block. Move the check ahead of those fetches.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — each step now runs under `tokio::time::timeout` (two refresh windows), so a wedged query leaves the step pending and retries instead of parking the loop forever. Worth noting the HTTP half was already covered — `fetch_json` and the scores scraper both set `HTTP_TIMEOUT_S` — so the real exposure was postgres and BigQuery.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — the guard now asks the DB (`SELECT EXISTS(SELECT 1 FROM validators)`) instead of the cold cache, so a full `validators` table with empty `cluster_info` stays 503 while a genuinely fresh environment still reaches ready. Verified locally in both directions: that state previously latched ready and served `total_count: 0`.

api/src/metrics.rs
ochaloup: 🤖 claude: Fixed — all six labels are seeded with 0 before the warm loop, so a cache that never loads exports a stale series rather than vanishing from the export. Verified with one step failing: its series was absent before and now reports 0 while the others carry real timestamps.

api/src/cache.rs
ochaloup: 🤖 claude: Declining — the check tests the output of `load_validators`, which consumes the overlays built from those very fetches, so it cannot be hoisted above them. Publishing the flags separately would break the invariant stated one line below, that flags and net APY publish with the records they stamped.

ochaloup:
@ilmari-h 

🤖 claude: `"cluster stats"` → `"cluster_stats"` applied, so the Prometheus label value no longer needs quoting in selectors. Added an assertion to `every_warm_step_is_named_and_distinct` that rejects whitespace in any step name, since the value ships as a label.

ilmari-h (approved):

PR #239, opened 2026-08-04
[GEN-8631] Historical take rates

ilmari-h:
Historical data for showing take rate history chart on the frontend per validator.

New cronjob that fills take rates and other validator rewards historical data in Postgres.

There was an existing BQ query to get take rates. Extended it so it serves the needs of the cache store as well as this PR and replaced it. Modified it to take absolute amounts instead of shares and moved the share calculation to the Rust code side.

## Summary by CodeRabbit

* **New Features**
  * Added take-rate collection, storage, and retrieval capabilities.
  * Added a `take-rates` command for collecting and storing validator reward data.
  * Added `GET /validators/{validator}/take-rates` with date or epoch filtering.
  * Take-rate results now include inflation, MEV, and block reward breakdowns.
* **Bug Fixes**
  * Improved validation and error handling for invalid or unavailable queries.
* **Documentation**
  * Added OpenAPI documentation for the take-rate endpoint and responses.


scripts/check-collect-store-take-rates.bash
```diff
@@ -0,0 +1,21 @@
...
+  echo "Collecting take rates"
+  # Pass-through args (e.g. --from-epoch N for a historical backfill, or --epochs-back N).
+  ${SCRIPT_DIR}/collect-take-rates.bash "$@" > "$output_file"
+  if [[ -s "$output_file" ]]; then
+    echo "Storing take rates from $output_file"
+    ${SCRIPT_DIR}/store-take-rates.bash "$output_file"
+  else
+    echo "No take rates data collected yet (BigQuery has no data for the window); skipping store, will retry next run"
+  fi
+  set +e
```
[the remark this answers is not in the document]
ilmari-h: same pattern in siblings, ignoring

collect/src/take_rates.rs
```diff
@@ -0,0 +1,227 @@
...
+        std::io::stdout(),
+        &TakeRatesSnapshot {
+            version: DATA_VERSION,
+            from_epoch,
+            loaded_at_epoch: current_epoch_info.epoch,
+            loaded_at_slot_index: current_epoch_info.slot_index,
+            created_at: created_at.to_string(),
+            take_rates,
+        },
+    )?;
```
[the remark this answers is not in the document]
ilmari-h: fixed

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,93 @@
...
+    tag = "Validators",
+    operation_id = "List take rate history",
+    path = "/validators/{vote_account}/take-rates",
+    params(
+        ("vote_account" = String, Path, description = "Vote account or identity of the validator"),
+        QueryParams
+    ),
+    responses(
+        (status = 200, body = ResponseTakeRates)
+    )
```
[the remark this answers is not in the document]
ilmari-h: fixed

collect/src/take_rates.rs
```diff
@@ -0,0 +1,242 @@
...
+        };
+
+        results.push(ValidatorEpochRewards {
+            epoch: u64_column(1, "epoch")?,
+            vote_account,
+            validator_rewards: u64_column(2, "validator_rewards")?,
+            total_rewards: u64_column(3, "total_rewards")?,
+            inflation_rewards: u64_column(4, "inflation_rewards")?,
+            mev_rewards: u64_column(5, "mev_rewards")?,
+            block_rewards: u64_column(6, "block_rewards")?,
```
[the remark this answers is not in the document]
ilmari-h: fixed

collect/src/take_rates.rs
```diff
@@ -0,0 +1,242 @@
...
+        solana_client_with_timeout(common_params.rpc_url, timeout, common_params.commitment);
+
+    let created_at = chrono::Utc::now();
+    let current_epoch_info = client.get_epoch_info()?;
+    info!("Current epoch: {current_epoch_info:?}");
+    let from_epoch = take_rates_params.from_epoch.unwrap_or_else(|| {
+        current_epoch_info
+            .epoch
+            .saturating_sub(take_rates_params.epochs_back)
+    });
```
ochaloup, line 218: 🤖 claude: The query is `epoch >= {from_epoch}` with no upper bound, so each run also writes the in-progress epoch and any freshly closed one whose MEV and Jito priority-fee passthrough have not landed in BigQuery yet — those rows credit the entire passthrough to the validator and read far above the real take rate.

`check/src/validators_block_rewards.rs` gates exactly this for the same tables (never the current epoch, wait 10 000 slots), so consider capping at `current_epoch - 1` and requiring `inflation_rewards > 0` as a completeness signal — `--epochs-back 2` can expire before Jito claims land, which makes the wrong value permanent rather than transient.

api/src/cache.rs
```diff
@@ -432,6 +438,29 @@ pub async fn warm_commissions_cache(context: &WrappedContext) -> anyhow::Result<
 
     Ok(())
 }
+pub async fn warm_take_rates_cache(context: &WrappedContext) -> anyhow::Result<()> {
+    info!("Loading take rates from DB");
+    let warmup_timer = Instant::now();
+    let take_rates = store::utils::load_take_rate_series(
+        &context.read().await.psql_client,
+        DEFAULT_CACHE_EPOCHS,
```
ochaloup: 🤖 claude: `DEFAULT_CACHE_EPOCHS` is 80 and the endpoint serves only from this cache, so every epoch a `--from-epoch` backfill writes below `last_epoch - 80` is stored, indexed and then unreachable — `query_from_date` silently returns a truncated series with no error.

Either a take-rate-specific constant here, or an explicit cap documented on the endpoint, so the frontend does not build an all-time control the API cannot satisfy.

collect/src/take_rates.rs
```diff
@@ -0,0 +1,242 @@
...
+                LEFT JOIN `{ds}.rewards_inflation` inflation
+                    ON stakes.stake_account = inflation.stake_account
+                    AND stakes.epoch = inflation.epoch AND inflation.epoch >= {from_epoch}
+                LEFT JOIN `{ds}.rewards_mev` mev
+                    ON stakes.stake_account = mev.stake_account
+                    AND stakes.epoch = mev.epoch AND mev.epoch >= {from_epoch}
+                -- rewards_validators_blocks is gross, so what Jito's PriorityFeeDistribution passed through has to come off the validator's keep rather than add to the pot.
+                LEFT JOIN `{ds}.rewards_jito_priority_fee` prio
+                    ON stakes.stake_account = prio.stake_account
+                    AND stakes.epoch = prio.epoch AND prio.epoch >= {from_epoch}
```
ochaloup, line 109: 🤖 claude: These three staker-side joins are not pre-aggregated, while all three validator-side joins below are — with the reason spelled out there as "so raw duplicate keys can't fan out the `SUM()`".

A second row for one `(stake_account, epoch)` in any of them, or in `stakes` for a mid-epoch re-delegation, double-counts the other two and understates the stored take rate. Cheap to settle: `SELECT stake_account, epoch, COUNT(*) FROM rewards_mev GROUP BY 1,2 HAVING COUNT(*) > 1 LIMIT 1`.

collect/src/take_rates.rs
```diff
@@ -0,0 +1,242 @@
...
+        help = "How many epochs back from the current epoch to (re-)query. Reward data can arrive a little late, so a small window is re-queried each run; already-stored epochs are idempotently upserted.",
+        default_value = "2"
+    )]
+    epochs_back: u64,
+
+    #[structopt(
+        long = "from-epoch",
+        help = "Query take rates from this epoch onwards. Overrides --epochs-back (use for historical backfill)."
+    )]
+    from_epoch: Option<u64>,
```
ochaloup, line 36: 🤖 claude: nit — `--from-epoch` is the documented backfill path but is unbounded and unchunked, so a low value scans every `stakes` partition three ways and emits one YAML that `store_take_rates` then loads whole into memory twice.

A `--to-epoch`, or a window loop in `check-collect-store-take-rates.bash`, would let a deep backfill run in batches.

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,94 @@
...
+                    if let Some(query_from_date) = query_params.query_from_date {
+                        take_rates = take_rates
+                            .iter()
+                            // Epochs with no boundaries yet can't be placed in the range.
+                            .filter(|v| {
+                                v.epoch_start_at
+                                    .is_some_and(|start_at| start_at > query_from_date)
+                            })
+                            .cloned()
+                            .collect();
```
ochaloup: 🤖 claude: An epoch with no `epochs` row has `epoch_start_at = None`, so this branch drops it while the `else` branch keeps it — the endpoint's two query modes return different newest-epoch sets.

`/commissions` does the opposite (it fills the null with `Utc::now()`, so it always includes that epoch). The nullable DTO here is the better design; worth stating the behaviour in the OpenAPI response description so the frontend knows which mode hides the open epoch.

migrations/0023-validators-rewards.sql
```diff
@@ -0,0 +1,22 @@
...
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
+
+  PRIMARY KEY(id),
+  UNIQUE(vote_account, epoch)
+);
+
+CREATE INDEX idx_validators_rewards_epoch
+    ON validators_rewards(epoch);
+CREATE INDEX idx_validators_rewards_vote_account
+    ON validators_rewards(vote_account);
```
ochaloup: 🤖 claude: nit — `UNIQUE(vote_account, epoch)` already builds a btree led by `vote_account`, so this index satisfies no read of its own and only adds write cost and disk.

The `epoch` index above it is not redundant and is the one `load_take_rate_series` needs — that one should stay.

store/src/utils.rs
```diff
@@ -505,6 +506,46 @@ pub async fn load_commissions(
     Ok(records)
 }
 
+pub async fn load_take_rate_series(
+    psql_client: &Client,
+    epochs: u64,
+) -> anyhow::Result<HashMap<String, Vec<TakeRateRecord>>> {
```
ochaloup: 🤖 claude: nit — this sits one screen from `load_take_rates`, which reads BigQuery and returns one scalar per validator for scoring, while this reads Postgres and returns a per-epoch series for the API cache.

Two names differing by one word for different stores, shapes and consumers; something like `load_take_rate_history_from_db` would take the coin flip out of picking one.

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,94 @@
...
+    query_params: QueryParams,
+    context: WrappedContext,
+) -> Result<impl Reply, warp::Rejection> {
+    info!("Fetching take rates {:?}", &vote_account);
+    metrics::REQUEST_COUNT_TAKE_RATES.inc();
+
+    let validators = context.read().await.cache.get_validators();
+    let validator = validators.iter().find(|(_vote_key, record)| {
+        record.identity == vote_account || record.vote_account == vote_account
+    });
```
ochaloup, line 60: 🤖 claude: nit — `get_validators()` returns the map by value, so this deep-clones every `ValidatorRecord` with its nested epoch stats on each request, just to run a linear find for one key — noticeable on a chart endpoint the frontend polls.

This exact block is now in five handlers (`commissions`, `versions`, `uptimes`, `events`, here); a `resolve_vote_key` helper borrowing under the read guard would remove the clone from all of them.

ochaloup (changes requested):
The general approach seems right to me. I would like ask to verify the Claude's findings as most of them seem relevant.

api/src/cache.rs
ilmari-h: I now fixed this so that we don't round-trip the take rate via cache. We read directly from postgres. The 80 limit disappears.

collect/src/take_rates.rs
ilmari-h: Using this pattern elsewhere and found it to be alright so keeping.

collect/src/take_rates.rs
ilmari-h: 🤖 BigQuery has no rows for the in-progress epoch, and an epoch's reward tables load together: I found no epoch in the last ~120 where block rewards landed without the matching MEV and priority-fee rows, and the amounts don't change after that load. So the partial state the cap would protect against doesn't occur, and re-querying later has nothing to correct.

collect/src/take_rates.rs
ilmari-h: 🤖 The data has no duplicate rows on either side, so the staker-side joins can't double-count and the validator-side subqueries aren't fixing anything. Adding subqueries to the staker side would make the query bigger and slower for zero change in output.

migrations/0023-validators-rewards.sql
ilmari-h: fixed

api/src/handlers/take_rates.rs
ilmari-h: fixed with clearer docs

store/src/utils.rs
ilmari-h: this was removed

api/src/handlers/take_rates.rs
ilmari-h: this seems to be a pre-existing pattern outside this pr so leaving for now

ilmari-h:
Thanks for the comments @ochaloup 

I decided to rewrite the implementation so that we remove the cache round-trip and the endpoint reads the data straight from PostgreSQL. So no cache-loading anymore. Not sure why I had that roundtrip there to begin with, it doesn't really serve a purpose, probably Claude's idea that I didn't question. Should be simpler now. Asking for another review.

api.md
```diff
@@ -8,10 +8,14 @@ Query parameters:
...
+- `order_field` - Default `Stake`, possible values: `Stake`, `Credits`, `MarinadeScore`, `Apy`, `NetApy`, `Commission`, `Uptime`, `TakeRate`, `ExpectedTakeRate`, `DelegationRelationships`, `Incidents`, and the group-only `Name`, `StakeDelta7d`, `StakeDelta30d`, `Validators` (these four order the `operators` array and leave the validators inside each operator on the `vote_account` tiebreak).
 - `order_direction` - Default `DESC`, possible values: `ASC`, `DESC`.
+- `with_operator_groups` - Optional, `true` also returns the `operators` array and groups the validators by operator: operators ordered by `order_field`, validators ordered by their operator's position and then by the same column. Validators with no operator come last. The array is never filtered and never paged, and its aggregates describe the whole eligible validator set rather than the filtered page.
 - `offset` - Default `0`.
-- `limit` - Default `100`.
+- `limit` - Default `100`. Pages `validators` only.
+
+Every validator carries `operator`, the name of its node operator from operators.csv, or `null` for
+a vote account the file does not list. It matches `operators[].name`, case aside.
```
ochaloup: 🤖 claude: None of `with_operator_groups`, the `operator` field, `operators.csv` or the six added `order_field` values exists anywhere in this repo — `OrderField` in `api/src/handlers/list_validators.rs:66` still has exactly the nine original variants. A client following these lines gets a serde failure on `order_field` and a silently ignored `with_operator_groups`.

```suggestion
- `order_field` - Default `Stake`, possible values: `Stake`, `Credits`, `MarinadeScore`, `Apy`, `NetApy`, `Commission`, `Uptime`, `TakeRate`, `ExpectedTakeRate`.
- `order_direction` - Default `DESC`, possible values: `ASC`, `DESC`.
- `offset` - Default `0`.
- `limit` - Default `100`.
```

api.md
```diff
@@ -138,6 +142,30 @@ curl -sfLS localhost:8000/validators/XkCriyrNwS3G4rzAXtG5B1nnvb5Ka1JtCku93VqeKAr
 }
 ```
 
+## Take rates
+Query parameters, mutually exclusive; without either one the whole stored history is returned:
+- `query_from_epoch` - Lower-bound epoch, inclusive.
+- `query_from_date` - Lower-bound RFC3339 date, resolved to the first epoch ending on/after it.
+
+`epoch_start_at` and `epoch_end_at` are `null` for epochs whose boundaries are not recorded, which
+also puts them out of reach of `query_from_date`.
```
ochaloup: 🤖 claude: `query_from_date` now resolves to an epoch number and filters `validators_rewards.epoch >= that`, so an epoch with no `epochs` row is still returned — `store/tests/take_rate_series_sql.rs` asserts exactly that for `LAST_EPOCH`. The frontend needs the null check on both query modes, not only the unbounded one.

```suggestion
`epoch_start_at` and `epoch_end_at` are `null` for epochs whose boundaries are not recorded.
Such an epoch is still returned by both query modes, including the open one.
```

store/src/take_rates.rs
```diff
@@ -0,0 +1,199 @@
...
+) -> anyhow::Result<()> {
+    info!("Storing take rates snapshot...");
+
+    let path = params.snapshot_path;
+    let snapshot_file = std::fs::File::open(&path)
+        .map_err(|e| anyhow::anyhow!("Failed to open snapshot take rates file '{path}': {e}"))?;
+    let snapshot: ValidatorRewardsSnapshot = serde_yaml::from_reader(snapshot_file)
+        .map_err(|e| anyhow::anyhow!("Failed to parse snapshot take rates file '{path}': {e}"))?;
+
+    let snapshot_created_at: DateTime<Utc> = snapshot.created_at.parse()?;
```
ochaloup, line 52: 🤖 claude: nit — `collect` writes `DATA_VERSION` into every snapshot but nothing ever reads it back, so a future layout change would be stored as if the format were unchanged. Either `anyhow::ensure!` the version here (which means making `DATA_VERSION` `pub`), or drop the field so it stops reading as a guarantee it does not give.

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,115 @@
...
+                    Err((
+                        StatusCode::INTERNAL_SERVER_ERROR,
+                        "Failed to fetch records!".into(),
+                    ))
+                }
+            },
+            (None, None) => Ok(None),
+        }
+    }
+}
```
ochaloup, line 30: 🤖 claude: nit — this `QueryParams` and `resolve_from_epoch` are a byte-for-byte copy of `api/src/handlers/events.rs:16-52`, doc comments and error strings included, so the two endpoints' `query_from_date` semantics can drift apart with no test on either side. A free `resolve_from_epoch(psql, query_from_epoch, query_from_date)` in `api/src/utils.rs` would leave each handler its own `IntoParams` struct plus a one-line delegation.

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,115 @@
...
+    path = "/validators/{vote_account}/take-rates",
+    params(
+        ("vote_account" = String, Path, description = "Vote account or identity of the validator"),
+        QueryParams
+    ),
+    responses(
+        (status = 200, body = ResponseTakeRates),
+        (status = 400, description = "Both query parameters given, or query_from_date outside the recorded epoch range"),
+        (status = 404, description = "No validator found for the given vote account or identity")
+    )
```
ochaloup, line 47: 🤖 claude: nit — the handler returns 500 on both error branches (`resolve_epoch_for_date` and `get_take_rate_series`), but the schema does not model it, so a generated client sees an unmodelled response instead of the documented error body. `events.rs:68`, which this was copied from, does list it.

```suggestion
    responses(
        (status = 200, body = ResponseTakeRates),
        (status = 400, description = "Both query parameters given, or query_from_date outside the recorded epoch range"),
        (status = 404, description = "No validator found for the given vote account or identity"),
        (status = 500, description = "Failed to fetch records")
    )
```

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,115 @@
...
+    let Some((vote_key, _validator)) = validator else {
+        error!("No validator found for {}", &vote_account);
+        return Ok(response_error(
+            StatusCode::NOT_FOUND,
+            "Failed to fetch records!".into(),
+        ));
+    };
+
+    let context_guard = context.read().await;
+    let psql_client = &context_guard.psql_client;
```
ochaloup, line 71: 🤖 claude: nit — `context_guard` lives to the end of the function, so the read guard spans both Postgres round-trips; `tokio::sync::RwLock` is fair, so a cache-warm `write()` queuing behind it then blocks every other reader for the duration. It matches `events.rs`, but this endpoint's no-parameter default returns the whole stored history, which a deep `--from-epoch` backfill is meant to make large.

store/tests/take_rate_series_sql.rs
```diff
@@ -0,0 +1,97 @@
...
+        .execute(
+            "INSERT INTO validators_rewards (
+                vote_account, epoch, validator_rewards, total_rewards, inflation_rewards,
+                mev_rewards, block_rewards, take_rate, created_at, updated_at
+            ) VALUES ($1, $2, 10, 100, 100, 0, 0, $3, NOW(), NOW())",
+            &[&vote_account, &Decimal::from(epoch), &take_rate],
+        )
+        .await
+        .unwrap();
+}
```
ochaloup, line 28: 🤖 claude: nit — the fixture writes rows with its own INSERT, so `store_take_rates` itself (the 10-array `UNNEST`, the `ON CONFLICT` update, the chunking) is never executed by any test; transposing two same-typed `NUMERIC` columns there would store a wrong reward mix silently and still pass. A second case that writes a snapshot YAML, runs `store_take_rates`, then reads it back through `get_take_rate_series` would cover it, plus the `created_at` RFC3339 round-trip.

ochaloup (approved):
It seems all much cleaner now, thank you.
Few nitpicks.

I'm moving to request a change but I would like to ask mostly only for aligning the `api.md` documentation. Other things are for your consideration. Feel free to just close them.

ochaloup (changes requested):
Sorry,  I wanted to ask for a change rather than to simply approve.

store/src/take_rates.rs
ilmari-h: done

api/src/handlers/take_rates.rs
ilmari-h: done

api.md
ilmari-h: done

api.md
ilmari-h: done

ilmari-h:
thanks @ochaloup

implemented a few of the nits I found worthwhile and aligned the api.md docs

api.md
ochaloup: 🤖 claude: Verified fixed — the `/validators` section is byte-identical to `master` again, so the whole `api.md` diff is now just the new take-rates section.

api.md
ochaloup: 🤖 claude: Verified fixed — the note now says what `store/tests/take_rate_series_sql.rs` actually asserts for the boundary-less epoch.

store/src/take_rates.rs
ochaloup: 🤖 claude: Verified fixed. One follow-on worth a thought, not blocking: the guard imports `DATA_VERSION` from `collect`, so reader and writer bump together and it cannot fire when a format change bumps both — a `store`-side `SUPPORTED_DATA_VERSION` is the form that catches that.

api/src/handlers/take_rates.rs
ochaloup: 🤖 claude: Verified fixed — one `resolve_from_epoch` in `api/src/utils.rs`, and folding `events.rs` in as well means the two endpoints can no longer drift apart.

store/src/take_rates.rs
```diff
@@ -0,0 +1,205 @@
...
+    let snapshot_file = std::fs::File::open(&path)
+        .map_err(|e| anyhow::anyhow!("Failed to open snapshot take rates file '{path}': {e}"))?;
+    let snapshot: ValidatorRewardsSnapshot = serde_yaml::from_reader(snapshot_file)
+        .map_err(|e| anyhow::anyhow!("Failed to parse snapshot take rates file '{path}': {e}"))?;
+
+    anyhow::ensure!(
+        snapshot.version == DATA_VERSION,
+        "Snapshot take rates file '{path}' has version {}, expected {DATA_VERSION}",
+        snapshot.version
+    );
```
ochaloup, line 50: 🤖 claude: nit — `DATA_VERSION` is imported from `collect`, so a future format change that bumps it recompiles both sides and this check becomes `2 == 2`, letting through exactly the mis-scaled rows it is meant to stop. A `SUPPORTED_DATA_VERSION` declared here in `store` instead would make the bump a conscious one.

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,91 @@
...
+    path = "/validators/{vote_account}/take-rates",
+    params(
+        ("vote_account" = String, Path, description = "Vote account or identity of the validator"),
+        QueryParams
+    ),
+    responses(
+        (status = 200, body = ResponseTakeRates),
+        (status = 400, description = "Both query parameters given, or query_from_date outside the recorded epoch range"),
+        (status = 404, description = "No validator found for the given vote account or identity")
+    )
```
ochaloup, line 47: 🤖 claude: nit — the handler returns 500 from two paths (`utils::resolve_from_epoch` and the `get_take_rate_series` branch below), so a client generated from this schema treats it as unmodelled and surfaces a parse error instead of the error body. `events.rs` lists it.

```suggestion
    responses(
        (status = 200, body = ResponseTakeRates),
        (status = 400, description = "Both query parameters given, or query_from_date outside the recorded epoch range"),
        (status = 404, description = "No validator found for the given vote account or identity"),
        (status = 500, description = "Failed to fetch records")
    )
```

api/src/handlers/take_rates.rs
```diff
@@ -0,0 +1,91 @@
...
+    let Some((vote_key, _validator)) = validator else {
+        error!("No validator found for {}", &vote_account);
+        return Ok(response_error(
+            StatusCode::NOT_FOUND,
+            "Failed to fetch records!".into(),
+        ));
+    };
+
+    let context_guard = context.read().await;
+    let psql_client = &context_guard.psql_client;
```
ochaloup, line 71: 🤖 claude: nit — this guard lives to the end of the function, so it spans both Postgres round-trips; `tokio::sync::RwLock` is fair, so the cache warmer's `context.write()` queues behind it and every reader arriving after that queues behind the writer. It bites harder here than in `events.rs` because this endpoint's default is the whole stored history rather than the last 80 epochs.

collect/src/take_rates.rs
```diff
@@ -0,0 +1,242 @@
...
+                CAST(SUM(COALESCE(vi.amount, 0) + COALESCE(vm.amount, 0)
+                    + GREATEST(COALESCE(vb.amount, 0) - staker_blocks, 0)) AS INT64) AS validator_rewards,
+                CAST(SUM(staker_inflation + COALESCE(vi.amount, 0)) AS INT64) AS inflation_rewards,
+                CAST(SUM(staker_mev + COALESCE(vm.amount, 0)) AS INT64) AS mev_rewards,
+                CAST(SUM(COALESCE(vb.amount, 0)) AS INT64) AS block_rewards,
+                CAST(SUM(staker_inflation + COALESCE(vi.amount, 0)
+                    + staker_mev + COALESCE(vm.amount, 0)
+                    + COALESCE(vb.amount, 0)) AS INT64) AS total_rewards
+            FROM stakers
+            -- Pre-aggregate to one row per (vote_account, epoch) so raw duplicate keys can't fan out the SUM().
```
ochaloup, line 126: 🤖 claude: nit — following on from the resolved thread about the staker-side joins: if duplicate keys cannot occur in this data, then this comment documents a hazard that was ruled out there. Worth rewording to say what the pre-aggregation is actually for, or dropping the claim.

ochaloup (dismissed):
It seem good to me. Claude has some nits so consider to involve or not.
If you do I would let this go to deployment in the next round.

store/src/take_rates.rs
ilmari-h: fixed

api/src/handlers/take_rates.rs
ilmari-h: fixed

ilmari-h:
@ochaloup thanks, 1 more commit to check

ochaloup (approved):
🚀

PR #240, opened 2026-08-05
[GEN-8006] [client id] adding a label

ochaloup:
Adding `client_label` to provide a similar experience to firedancer.io 

## Summary by CodeRabbit

* **New Features**
  * Added standardized client labels to validator, version, and client metadata.
  * Client labels now resolve from registered client IDs, reported names, or an “Unknown” fallback.
  * Added support for distinguishing client vendors and implementation lineages.

* **Bug Fixes**
  * Preserved nullable client names while consistently generating display labels.
  * Improved handling of named, numeric, and missing client identifiers.


store/tests/store_client_columns_sql.rs
```diff
@@ -331,27 +331,43 @@ async fn load_versions_serves_unknown_when_no_client_name_is_stored() {
...
-        vec!["Agave".to_string(), UNKNOWN_CLIENT_NAME.to_string()],
+        vec![
+            "Agave".to_string(),
+            "FireBAM".to_string(),
+            UNKNOWN_CLIENT_NAME.to_string()
+        ],
         "a NULL client_name reads back as the fallback, a stored one unchanged"
     );
 
+    let mut labels: Vec<String> = records.iter().map(|r| r.client_label.clone()).collect();
```
honzsik, line 361: This is the only `client_label` assertion, and it exercises the `load_versions` path. The `load_validators` path derives the label at two separate sites in `store/src/utils.rs` (the `ValidatorRecord` default and the `label.clone()` into `ValidatorEpochStats`), and neither is covered — this test file never calls `load_validators`.

On #238 the request was to assert the `validators` and `versions` read paths independently rather than assume they behave the same, and that was accepted. The same reasoning applies to `client_label` here: a regression in the `load_validators` label computation (or a mismatch between the two sites) would pass every test in this file. Consider a parallel `load_validators` case asserting `client_label` on both the record and its epoch stats.

honzsik (changes requested):
Requesting changes on 1 item.

1. `client_label` asserted only on the `load_versions` path, not `load_validators` — store/tests/store_client_columns_sql.rs:359

store/tests/store_client_columns_sql.rs
ochaloup: I updated the code, thanks for the review.

ochaloup:
@honzsik I updated the PR, can you please re-review?

store/tests/store_client_columns_sql.rs
```diff
@@ -331,27 +333,139 @@ async fn load_versions_serves_unknown_when_no_client_name_is_stored() {
...
+    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let port = listener.local_addr().unwrap().port();
+    tokio::spawn(async move {
+        let (mut socket, _) = listener.accept().await.unwrap();
+        let mut request = [0u8; 1024];
+        let read = socket.read(&mut request).await.unwrap();
+        assert!(
+            String::from_utf8_lossy(&request[..read]).contains("/validators/verified"),
+            "the stub answers the verified-validators endpoint only"
+        );
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — the stub now wraps the socket in a `BufReader` and reads the full request line with `read_until` before asserting on `/validators/verified`.

ochaloup:
🤖 claude: Declining the docstring-coverage pre-merge warning — this repo deliberately keeps doc comments off these modules, so reaching an 80% threshold would mean adding `///` blocks that only restate the names.

store/src/dto.rs
```diff
@@ -12,6 +13,16 @@ use std::collections::HashMap;
 /// Served instead of null so every consumer has a client name to render.
 pub const UNKNOWN_CLIENT_NAME: &str = "Unknown";
 
+// Derived on read rather than stored: the label is a function of client_id alone, so a column would
```
honzsik: The comment says the label "is a function of client_id alone," but the signature takes `reported_name` too and the body falls back to it: `client_id.and_then(|id| ClientId::Registered(id).label()).or(reported_name)`. A NULL or unresolved `client_id` with a reported name of `Agave` yields the label `Agave` — the `load_versions`/`load_validators` tests in this PR assert exactly that. So the label is a function of `client_id` *and* the reported name.

The justification for deriving-on-read is still sound, but the premise as written is inaccurate on a newly-added line, and a future refactor could trust it and drop the second parameter. Consider rewording to something like "the label is a function of `client_id` (with the reported name as fallback)."

store/tests/store_client_columns_sql.rs
```diff
@@ -331,27 +333,140 @@ async fn load_versions_serves_unknown_when_no_client_name_is_stored() {
...
+        .batch_execute(&format!("DROP SCHEMA {schema} CASCADE"))
+        .await
+        .unwrap();
+}
+
+// `load_validators` aborts if the validator-bonds API is unreachable, so it must be answered here.
+async fn verified_validators_stub() -> String {
+    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let port = listener.local_addr().unwrap().port();
+    tokio::spawn(async move {
```
honzsik, line 383: Optional, diagnosability only. The spawned task's `JoinHandle` is discarded, so if the `assert!(... "/validators/verified" ...)` or any `.unwrap()` in here ever fires, the task panics silently and the test fails later at the outer `load_validators(...).await.unwrap()` with an opaque connection-reset error instead of this assertion's message. A future change to `load_validators`'s request shape would still fail the test, but the cause would be much harder to read from CI logs. The exercised path is correct today, and this mirrors the fire-and-forget spawn idiom already in `store/tests/common/mod.rs`, so leaving it is defensible.

honzsik (commented):
Round-1 comment on the `load_validators` label path is addressed — the new test calls `load_validators` and asserts `client_label` on both the record and its epoch stats. Three follow-up notes below; only the first is worth acting on, 2 and 3 are optional.

1. `client_label` comment says the label is "a function of client_id alone" but it also falls back to `reported_name` — store/src/dto.rs:16
2. (optional) stub's spawned task discards its JoinHandle, so an endpoint-assert failure surfaces as an opaque transport error — store/tests/store_client_columns_sql.rs:383
3. (optional) id 0 (solana-labs) collapses to bare `Agave` despite vendor != lineage, undocumented by "when there is one" — collect/src/solana_service.rs:202

collect/src/solana_service.rs
```diff
@@ -178,34 +178,41 @@ impl ClientId {
...
-            5 => ("firedancer", "firedancer"),
-            6 => ("bam", "agave"),
-            7 => ("sig", "sig"),
-            8 => ("rakurai", "agave"),
-            9 => ("harmonic", "firedancer"),
-            10 => ("harmonic", "agave"),
-            11 => ("harmonic", "frankendancer"),
-            12 => ("bam", "frankendancer"),
-            13 => ("raiku", "agave"),
+            0 => ("solana-labs", "agave", "Agave"),
```
honzsik, line 203: Optional, documentation clarity. The label is described as "the lineage, plus the vendor's modification of it when there is one." Every id where vendor != lineage gets a `+ X` suffix (jito, paladin, bam, rakurai, harmonic, raiku) — except id 0 `("solana-labs", "agave", "Agave")`, which stays the bare `Agave` and is indistinguishable from id 3 `("agave", "agave", "Agave")` at the label level. ids 2 and 5 earned an explanatory comment for their deviation; id 0's silent collapse could use one too, so a reader can tell it's deliberate legacy handling rather than a missing arm. `client_vendor` still distinguishes the two, so there's no functional issue.

store/src/dto.rs
ochaloup: Ok

store/tests/store_client_columns_sql.rs
ochaloup: I would not do this

collect/src/solana_service.rs
ochaloup: Ok

store/src/dto.rs
ochaloup: 🤖 claude: Fixed — the comment now says the label follows `client_id` with the reported name as fallback, since that branch is exactly what the `load_versions`/`load_validators` label tests assert.

store/tests/store_client_columns_sql.rs
ochaloup: 🤖 claude: Declining — retaining the `JoinHandle` means changing the stub's return type and awaiting it in the test, which is more scaffolding than the diagnostic is worth given the test fails either way.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Added a line at the id 0 arm noting it is Agave's pre-rename vendor rather than a fork of it, so the bare label reads as deliberate.

ochaloup:
@honzsik PR updated

honzsik (approved):
The client_label comment and the id 0 labeling note are both clarified in 0f0b4d1, and the optional JoinHandle point is fairly declined. The load_validators test now covers the label on both the record and its epoch stats, closing the round-1 gap. Reviewed the new commits against git history and the surrounding client-id machinery; no further issues.

PR #241, opened 2026-08-06
[GEN-8638] client data handling unification

ochaloup:
A follow-up to [GEN-8006]: https://github.com/marinade-finance/delegation-strategy-2/pull/217

Client name, vendor and lineage become read-time derivations of `client_id` rather than stored columns, so adding a `collect/client-ids.csv` row reclassifies every epoch already collected instead of only rows written after the deploy.

- Drop the three columns in migration `0020` — **deploy the binaries first**, since the running release still reads them and a dropped column cannot be rolled back
- Re-resolve `client_id_raw` on read when no numeric id was stored, so clients first seen before their registry row existed reclassify too rather than staying unknown forever
- Serve `Unknown` as the name of a client the registry does not know, keeping the node's exact gossip string on `client_id_raw`
- Gate all four derived fields on the registry knowing both the name and the vendor/lineage mapping, so a record cannot read as half-classified
- Fold the per-client-id stake distribution into vendor and lineage buckets in Rust, merging the several ids one vendor ships, and build `/cluster-stats` from one query instead of two identical ones
- Leave every API field's name and type unchanged

## Summary by CodeRabbit

* **Improvements**
  * Client information is now consistently resolved from registered client IDs.
  * Raw client identifiers remain available for classification and display when needed.
  * Unrecognized clients receive clear fallback labeling.
  * Validator, version, and cluster statistics now use consistent client identification and grouping.
  * Client diversity and lineage statistics better combine equivalent reported client formats.

* **Data Updates**
  * Removed legacy derived client metadata from stored validator and version records.


ochaloup:
🤖 claude:

**`agg_versions` tiebreaker — fixed.** Added `id desc` after `created_at desc` on all three aggregates, matching `store/src/versions.rs`, so the shared-row read that `effective_client_id` depends on is guaranteed rather than incidental. One correction to the finding: a tie is not reachable via `store_versions`, since `snapshot.validators` is keyed by vote_account and contributes at most one row per vote_account per batch — it would take two snapshots with an identical nanosecond `created_at`. Worth doing as hardening regardless.

**Docstring coverage 54% vs 80% — declining.** Rust docstrings are `///`, which this project's comment policy bans outright; the threshold belongs in the CodeRabbit org config rather than in a dozen new doc comments here.

ilmari-h (approved):
Looks good!

store/src/utils.rs
```diff
@@ -1808,7 +1861,7 @@ pub async fn load_validators_aggregated_flat(
...
-                agg_versions AS (select vote_account, (array_agg(version order by created_at desc) filter (where version is not null))[1] as last_version, (array_agg(client_vendor order by created_at desc) filter (where client_vendor is not null and epoch <= $2))[1] as last_client_vendor, (array_agg(client_lineage order by created_at desc) filter (where client_lineage is not null and epoch <= $2))[1] as last_client_lineage from versions group by vote_account)
+                agg_versions AS (select vote_account, (array_agg(version order by created_at desc, id desc) filter (where version is not null))[1] as last_version, (array_agg(client_id order by created_at desc, id desc) filter (where (client_id is not null or client_id_raw is not null) and epoch <= $2))[1] as last_client_id, (array_agg(client_id_raw order by created_at desc, id desc) filter (where (client_id is not null or client_id_raw is not null) and epoch <= $2))[1] as last_client_id_raw from versions group by vote_account)
```
ilmari-h, line 1864: ### 🤖 last_client_vendor / last_client_lineage semantics changed for scoring input
Before: last **non-null vendor** row won, so an unclassified newer reading fell through to the last known vendor. Now the filter is `(client_id is not null or client_id_raw is not null)`, so the newest row that reported *any* client wins even when unresolvable, and the validator drops to `unknown`.

`ValidatorAggregatedFlat.client_vendor/client_lineage` feeds scoring. A validator that switches to a client not yet in `client-ids.csv` now loses its previous client classification instead of keeping it. Test `validators_flat_takes_id_and_rendering_from_the_same_versions_row` asserts this is intended — confirm with whoever owns the scoring weights before merge.

ilmari-h (commented):
just 1 finding from Claude, not sure how relevant

ilmari-h (approved):
approved with 1 claude nitpick

store/src/utils.rs
ochaloup: Thanks for the point, but I will skip fixing this.

---

validators_flat_reports_unknown_for_an_unregistered_numeric_client_over_an_older_classified_row — same shape as validators_flat_takes_id_and_rendering_from_the_same_versions_row (older row Some(5)/Some("Firedancer"), newer row None/Some("Unknown(86)")), asserting client_vendor and client_lineage are both
"unknown".

Why this one: the name-rendering path (Raiku2 → Unrecognized(None)) is already pinned by two flat tests, but the numeric path (Unknown(86) → Unrecognized(Some(86))) only has a resolve_client_id unit test in collect/src/solana_service.rs and never reaches load_validators_aggregated_flat.

PR #242, opened 2026-08-06
[GEN-8245][api] adding bonds protected field to listing validators in similar fashion as verified does

ochaloup:
Loading the `/protected` bonds endpoint from https://github.com/marinade-finance/validator-bonds/pull/489
to validators data in a similar manner as `/filtered` was done.
Per discussion at https://marinade-group.slack.com/archives/C0BJZAZC66S/p1786007415636699?thread_ts=1786007411.934379&cid=C0BJZAZC66S

---

Adds a `protected` flag and `query_protected` filter to `/validators`, unblocking server-side pagination of the Explore list.

- Add `protected` per validator and a `query_protected` filter, sourced from validator-bonds `GET /validators/protected`, so Explore can drop its client-side bond download
- Move both bond-flag fetches into the cache warmer behind one shared loader, so `verified` and `protected` share one guard and one failure policy instead of two divergent copies
- Reuse the last known list when the bonds API fails, bounded to an hour, so a brief outage stops freezing every validator field while a permanent one still surfaces
- Reject an empty list from either endpoint, since clearing the flag on every validator is never a correct answer to a successful request

## New Features
- Added support for identifying protected and verified validators.
- Added filtering for protected, unprotected, or all validators.
- Validator records now include protected status alongside delegator counts and commission rates.
- Validator responses now include bond-status freshness information.

## Bug Fixes
- Improved validator data refresh reliability by temporarily preserving known bond statuses during upstream failures, with bounded fallback reuse.
- Improved handling of empty responses, cold starts, and failed data requests.
- Improved consistency of staking and validator data processing.
- Added coverage for fallback limits and protected-validator behavior.

### DIFF
--- a/api/src/cache.rs
+++ b/api/src/cache.rs
@@ -1,13 +1,15 @@
 use crate::context::WrappedContext;
 use log::{error, info};
 use rust_decimal::Decimal;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::time::{SystemTime, UNIX_EPOCH};
 use store::dto::{
     ClusterStats, CommissionRecord, ScoringRunRecord, UptimeRecord, ValidatorRecord,
     ValidatorScoreRecord, VersionRecord,
 };
 use tokio::time::{sleep, Duration, Instant};
+
+use store::utils::ValidatorOverlays;
 
 pub(crate) use store::utils::DEFAULT_CACHE_EPOCHS;
 pub(crate) const DEFAULT_COMPUTING_EPOCHS: u64 = 20;
@@ -31,8 +33,22 @@ pub struct CachedMultiRunScores {
     pub scores: HashMap<Decimal, Vec<ValidatorScoreRecord>>,
 }
 
+/// One validator-bonds flag list plus how many refreshes in a row have fallen back to it.
+#[derive(Default, Clone)]
+pub struct CachedFlag {
+    pub vote_accounts: HashSet<String>,
+    pub consecutive_failures: u32,
+}
+
+#[derive(Default, Clone)]
+pub struct CachedBondFlags {
+    pub verified: CachedFlag,
+    pub protected: CachedFlag,
+}
+
 #[derive(Default)]
 pub struct Cache {
+    pub bond_flags: CachedBondFlags,
     pub validators: CachedValidators,
     pub commissions: CachedCommissions,
     pub versions: CachedVersions,
@@ -165,6 +181,38 @@ impl Cache {
     }
 }
 
+/// Bounded so a permanently broken bonds API surfaces as an outage instead of flags frozen forever.
+const MAX_FLAG_FALLBACK_CYCLES: u32 = 6;
+
+fn resolve_flag(
+    flag: &str,
+    fetched: anyhow::Result<HashSet<String>>,
+    last: CachedFlag,
+) -> anyhow::Result<CachedFlag> {
+    let err = match fetched {
+        Ok(vote_accounts) => {
+            return Ok(CachedFlag {
+                vote_accounts,
+                consecutive_failures: 0,
+            })
+        }
+        Err(err) => err,
+    };
+
+    let consecutive_failures = last.consecutive_failures + 1;
+    if last.vote_accounts.is_empty() || consecutive_failures >= MAX_FLAG_FALLBACK_CYCLES {
+        return Err(err);
+    }
+
+    error!(
+        "Failed to load {flag} validators, reusing the last known list ({consecutive_failures}/{MAX_FLAG_FALLBACK_CYCLES}): {err}"
+    );
+    Ok(CachedFlag {
+        consecutive_failures,
+        ..last
+    })
+}
+
 pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<()> {
     info!("Loading validators from DB");
     let warmup_timer = Instant::now();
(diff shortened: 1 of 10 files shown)


api/src/handlers/list_validators.rs
```diff
@@ -47,6 +47,8 @@ pub struct QueryParams {
     query_sfdp: Option<bool>,
     query_incident_free: Option<bool>,
     query_verified: Option<bool>,
+    /// Validators whose bond can back a PSR downtime claim, as reported by validator-bonds. A presence check, not a sufficiency check — see the `protected` field.
```
kronael: others dont have a comment and are fine

api/src/handlers/reports_staking.rs
```diff
@@ -66,8 +66,7 @@ async fn get_planned_stakes(context: WrappedContext) -> anyhow::Result<Vec<Staki
                 let current_epoch_stats = validator
                     .epoch_stats
                     .iter()
-                    .filter(|validator| validator.epoch == last_epoch)
-                    .next_back();
+                    .rfind(|validator| validator.epoch == last_epoch);
```
kronael, line 69: love that I learn new things

api/src/cache.rs
```diff
@@ -31,8 +33,22 @@ pub struct CachedMultiRunScores {
     pub scores: HashMap<Decimal, Vec<ValidatorScoreRecord>>,
 }
 
+/// One validator-bonds flag list plus how many refreshes in a row have fallen back to it.
+#[derive(Default, Clone)]
+pub struct CachedFlag {
+    pub vote_accounts: HashSet<String>,
+    pub consecutive_failures: u32,
```
kronael: based on the commnet it would seem to me these are not "error" states which I would assume just by seeing "failures"

api/src/cache.rs
```diff
@@ -165,6 +181,38 @@ impl Cache {
     }
 }
 
+/// Bounded so a permanently broken bonds API surfaces as an outage instead of flags frozen forever.
+const MAX_FLAG_FALLBACK_CYCLES: u32 = 6;
+
+fn resolve_flag(
```
kronael, line 195: tohught: It would seem to me that this might be a responsibility of the user, or perhaps "marinade-graph"?  (but its not there yet so possibly just thinking about it)

api/src/cache.rs
```diff
@@ -178,14 +226,37 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
         .map(|c| (c.unique_delegators.clone(), c.take_rates.clone()))
         .unwrap_or_default();
 
+    let scoring_url = context.read().await.scoring_url.clone();
```
kronael: It would seem you'd want to do just one lock and not three here.

collect/src/solana_service.rs
kronael, line 797: nit: it would seem using anyhow would be more fitting

kronael (dismissed):
just some nits

api/src/cache.rs
ochaloup: 🤖 claude: Keeping it in DS for now — the fallback is bounded to 6 refresh cycles and then fails loudly, so it cannot become a permanent DS-side fiction. Moving the staleness decision to the consumer needs a staleness signal in the response first, which does not exist yet.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Declining — that signature is pre-existing on master, and the Box is what satisfies clippy result_large_err. anyhow is already applied one frame out, where the error is unboxed and given context.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — the three reads are now one scoped read guard returning a tuple, kept in an explicit block so it is released before the awaits that follow.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — renamed to consecutive_fallbacks, since the counter only increments on the successful degradation path and stops the moment the refresh actually errors.

api/src/handlers/list_validators.rs
ochaloup: 🤖 claude: Fixed — comment removed to match the neighbouring filters. The semantics stay documented on ValidatorRecord::protected, which is where the field itself is described.

api/src/cache.rs
```diff
@@ -165,6 +181,43 @@ impl Cache {
...
+        Ok(vote_accounts) => {
+            if vote_accounts.is_empty() {
+                error!(
+                    "The {flag} endpoint lists nobody, so the flag is cleared on every validator"
+                );
+            }
+            return Ok(CachedFlag {
+                vote_accounts,
+                consecutive_fallbacks: 0,
+            });
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Addressed, though not as a straight revert — an empty list now reuses the last known list and counts against the same bound a failure gets, and is only believed once it survives that bound. One malformed empty response can no longer clear the flag, while a genuinely empty upstream still propagates instead of being permanently unrepresentable. Note the premise about a pre-existing rule is off: the emptiness rejection was introduced by this PR and never existed on master, which accepts whatever list comes back.

api/src/cache.rs
```diff
@@ -178,14 +231,43 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
...
     let validators = store::utils::load_validators(
         &context.read().await.psql_client,
-        context.read().await.scoring_url.clone(),
-        context.read().await.validator_bonds_api_url.clone(),
+        scoring_url,
         DEFAULT_CACHE_EPOCHS,
         DEFAULT_COMPUTING_EPOCHS,
-        &unique_delegators,
-        &take_rates,
+        &overlays,
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining here — the guard is pre-existing and identical on master, and this branch shortens the guarded window rather than widening it by moving both bonds HTTP calls out of it. Arc<Client> also will not compile, since store_scoring takes &mut Client for its transaction, so the real fix is a connection pool and belongs in its own change.

kronael (approved):

PR #244, opened 2026-08-07
[GEN-8658] fix commission bps

ochaloup:
Agave 4.x updated the getInflationReward RPC call to stop providing data in the JSON field commission and started providing it in commissionBps instead (https://github.com/anza-xyz/agave/blob/v4.2.0-rc.1/runtime/src/inflation_rewards/points.rs#L59-L61). When using that method, commissionBps needs to be used (https://app.notion.com/p/3b5e465715a4812393c4ee23a3cc3e37)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana inflation reward processing across legacy and newer response formats.
  * Commission rates are interpreted more accurately, including basis-point values, rounding, and limits up to 100%.
  * Added safer handling for missing or conflicting commission data to prevent inaccurate reward calculations.
  * Inflation reward requests now honor configured commitment and epoch settings for more consistent results.
  * Improved reporting of commission-rate resolution details when source data is incomplete or inconsistent.

### DIFF
--- a/collect/src/solana_service.rs
+++ b/collect/src/solana_service.rs
@@ -6,13 +6,13 @@ use bincode::deserialize;
 use log::{info, warn};
 use rust_decimal::{prelude::ToPrimitive, Decimal};
 use serde::Deserialize;
-use serde_json::{Map, Value};
+use serde_json::{json, Map, Value};
 use solana_account_decoder::validator_info;
 use solana_account_decoder::UiAccountEncoding;
 use solana_client::{
     client_error::ClientError,
     rpc_client::RpcClient,
-    rpc_config::{RpcAccountInfoConfig, RpcProgramAccountsConfig},
+    rpc_config::{RpcAccountInfoConfig, RpcEpochConfig, RpcProgramAccountsConfig},
     rpc_filter::{Memcmp, RpcFilterType},
     rpc_request::RpcRequest,
     rpc_response::RpcVoteAccountStatus,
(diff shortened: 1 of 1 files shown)


collect/src/solana_service.rs
```diff
@@ -614,6 +614,61 @@ pub fn get_withdraw_authorities(
...
+    lossy: usize,
+    disagree: usize,
+    unresolved: usize,
+    no_reward: usize,
+}
+
+// Agave projects commissionBps onto the legacy percent this way; rounding down instead would let a
+// validator above the 10% eligibility cap read as exactly at it.
+fn bps_to_percent(bps: u16) -> u8 {
+    bps.min(10_000).div_ceil(100) as u8
```
martinkrecek, line 642: basic-data-collector PR 42 rounds down instead, shouldn't both services have the same approach?

collect/src/solana_service.rs
```diff
@@ -626,21 +681,66 @@ pub fn get_commission_from_inflation_rewards(
...
+    if result.len() * 2 < queried {
+        warn!(
+            "Resolved commission for {} of {} validators: {} without a reward, {} served neither commission nor commissionBps",
+            result.len(),
+            queried,
+            stats.no_reward,
+            stats.unresolved
+        );
+    } else {
+        info!(
```
martinkrecek: (non-blocking)
I'd consider logging this info every time, not only in the else branch.

martinkrecek (dismissed):

collect/src/solana_service.rs
martinkrecek: confirmed in the other PR - ceil is correct

ochaloup: Decided to use `ceil` in both

martinkrecek (approved):

PR #245, opened 2026-08-10
[GEN-8656][api] adding min downtime seconds

ochaloup:
Adds a downtime floor to the `/validators` incident-free filter, so the Explore list can stop narrowing it client-side and let it compose with pagination.

- Add `min_downtime_seconds` so `query_incident_free` ignores `DOWN` intervals shorter than the floor, matching the 3-minute rule the Explore column and the validator detail page already apply
- Default the floor to 0, so existing callers of `query_incident_free` get exactly today's results
- Leave the returned `incidents` array unfiltered, since the detail page reads it from the same endpoint and a param-dependent payload would let the two surfaces drift apart again
- Pin the `>=` boundary in tests — 179s kept, 180s dropped — alongside the inverted direction and the param used on its own

## New Features
- Added an optional minimum incident-downtime threshold for validator queries and configuration.
- Validator results can now be filtered based on downtime across all recorded incidents.
- The threshold defaults to zero and remains inactive when used without other incident filters.

## Bug Fixes
- Improved incident filtering across multiple incidents and query conditions.

## Documentation
- Clarified how validator protection status is determined.

### DIFF
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -47,7 +47,10 @@ pub struct QueryParams {
     query_marinade_stake: Option<bool>,
     query_with_names: Option<bool>,
     query_sfdp: Option<bool>,
+    /// Evaluated over the last 90 epochs of incidents, regardless of `epochs` and `query_from_date`.
     query_incident_free: Option<bool>,
+    /// Minimum downtime in seconds for a `DOWN` interval to count as an incident for `query_incident_free`. Shorter intervals are restart noise. Ignored unless `query_incident_free` is set, and never filters the returned `incidents` array.
+    min_downtime_seconds: Option<u64>,
     query_verified: Option<bool>,
     query_protected: Option<bool>,
     query_flagged: Option<bool>,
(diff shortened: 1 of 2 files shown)


api/src/handlers/list_validators.rs
```diff
@@ -47,7 +47,10 @@ pub struct QueryParams {
     query_marinade_stake: Option<bool>,
     query_with_names: Option<bool>,
     query_sfdp: Option<bool>,
+    /// Evaluated over the last 90 epochs of incidents, regardless of `epochs` and `query_from_date`.
     query_incident_free: Option<bool>,
+    /// Minimum downtime in seconds for a `DOWN` interval to count as an incident for `query_incident_free`. Shorter intervals are restart noise. Ignored unless `query_incident_free` is set, and never filters the returned `incidents` array.
+    min_downtime_seconds: Option<u64>,
```
martinkrecek: (non-blocking - minor)
Since this new param and `query_incident_free` are tied together, I'd consider naming this one like `min_incident_downtime_seconds` or similar to flag it's related to that filter. The comment already explains pretty clearly, though, so this is not really necessary and feel free to resolve.

martinkrecek (dismissed):

ochaloup:
@martinkrecek I like the renaming idea so I changed the code. Can you reviewe please? :pray:

martinkrecek (approved):

PR #246, opened 2026-08-10
[store] call the validator-bonds flag endpoints under /v1

ochaloup:
I need to make changes to how validator bonds define a protected validator (https://marinade-group.slack.com/archives/C0BJZAZC66S/p1786171658277399?thread_ts=1786007411.934379&cid=C0BJZAZC66S)
With that, I want to migrate to /v1, as we usually use in marinade.

Both /validators/verified and /validators/protected move behind the /v1 prefix. Merge only after validator-bonds serves /v1 in production.

**REQUIRED** to first deploy changes from https://github.com/marinade-finance/delegation-strategy-2/pull/246#issue-5110700329

### DIFF
--- a/store/src/utils.rs
+++ b/store/src/utils.rs
@@ -798,7 +798,7 @@ pub async fn load_protected_validators(base: &str) -> anyhow::Result<HashSet<Str
     .await
 }
 
-// `base` is the validator-bonds API base URL; `/validators/{flag}` is appended here.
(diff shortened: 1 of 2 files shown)


martinkrecek (approved):
We could start a philosophical discussion about whether `v1/` belongs in the base URL (all latest endpoints share the same version and clients just pick a version) or not (endpoints bump their versions individually and clients pick per endpoint), but this is not the right place and it's not important enough, I'd say. 🙂

PR #250, opened 2026-08-12
[GEN-8660][SIMD-0525] fixing the hardcoded 400ms slot time

ochaloup:
Replace the hardcoded 400ms slot assumption with values read from the chain, so SIMD-0525's staged slot-time reductions do not distort the SAM auction or the collectors.

- Read `slots_per_year` from the SIMD-0525 feature gates and store it per epoch, so the auction's inflation input tracks the protocol instead of a fixed 182.625 epochs/year
- Backfill existing epochs with agave's 400ms baseline so no historical APY or auction figure moves, pinned to 0.01% by an integration test
- Gate block-rewards collection on wall clock rather than slot counts, anchored on a block the node still holds so a pruned ledger cannot silently disable the check
- Publish the per-epoch nominal on `/rewards`, including the running epoch, so ds-sam can normalise its rewards window across a slot-time change
- Correct `inflation` and `supply` to the epoch being closed, since the finalise job backfills past epochs where the live RPC values describe today
- Warn when the measured slot time belongs to a different row of the gate table than the selected one, since a mistyped gate pubkey otherwise reads exactly like "not activated"
- Delete the unused `get_apy` and its hardcoded `epochs_in_year = 160`

## New Features
- Added epoch-specific slot-rate and annual slot estimates.
- Rewards and performance data now include the slot-rate assumptions used for calculations.
- Added support for protocol slot-time changes activated across epochs.

## Improvements
- Inflation, APR, and APY calculations now reflect applicable slot rates.
- Wait-time estimates use measured network timing when available.
- Historical and running-epoch reward estimates preserve slot-rate information.

## Bug Fixes
- Improved handling of skipped, pruned, or insufficiently timed blocks.
- Added safer fallbacks when network timing cannot be measured.


collect/src/slot_params.rs
```diff
@@ -0,0 +1,212 @@
...
+pub const SLOTS_IN_EPOCH: u64 = 432_000;
+
+/// Agave's pre-SIMD-0525 baseline: no gate governs it, so no account can be read for it.
+const BASELINE_SLOT_PARAMS: SlotParams = SlotParams {
+    slot_time_ms: 400,
+    slots_per_year: 78_892_314.984,
+};
+
+/// Transcribed from agave `runtime/src/slot_params.rs`, which stores explicit rows rather than deriving them.
+const SLOT_TIME_GATES: [(Pubkey, SlotParams); 4] = [
```
ochaloup, line 16: I'm not super happy about this, but I didn't want to do a full version update of everything.

collect/src/validators_performance.rs
```diff
@@ -226,16 +310,13 @@ pub fn collect_validators_performance_info(
         None
     };
 
+    let slots_per_year = get_slot_params(&client, epoch)?.slots_per_year;
```
ilmari-h: ### 🤖 The safety net sits behind an `if`; the thing it protects does not** 

```rust
let slots_per_year = get_slot_params(&client, epoch)?.slots_per_year;   // 313 — ALWAYS runs

let cluster_inflation = if performance_params.with_rewards {            // 315
    warn_on_slot_time_divergence(&client, &current_epoch_info);         // 316 — only sometimes
    ...
};
```

A mistyped gate pubkey returns the 400ms baseline — identical to a healthy cluster today, undetectable
on chain. The clock comparison is the only detector, as `:107` says.

But `slots_per_year` is written out on every run (snapshot -> `cluster_info.slots_per_year` -> served
by `/rewards`), while the check runs only under `with_rewards`. A `--with-rewards`-off run therefore
persists and serves a possibly wrong regime with the alarm disarmed.

Hoist the call up next to line 313, outside the `if`.

collect/src/validators_performance.rs
```diff
@@ -91,11 +96,90 @@ pub struct ValidatorsPerformanceSnapshot {
...
+    // so walking it backwards is bounded by `initial` at one end and pinned by `terminal` at the other.
+    let inflation = if inflation_now <= terminal {
+        inflation_now
+    } else {
+        (inflation_now * (1.0 - taper).powf(-(epochs_behind as f64) / nominal_epochs_per_year))
+            .min(initial)
+    };
+    let minted_since = (1.0 + inflation_now / nominal_epochs_per_year).powi(epochs_behind as i32);
+
+    (inflation, (supply_now as f64 / minted_since) as u64)
```
ilmari-h: ### 🤖 The supply back-walk uses today's inflation rate while the rate itself is tapered** —
`collect/src/validators_performance.rs:172-180`

```rust
let inflation = ... inflation_now * (1.0 - taper).powf(-(epochs_behind as f64) / ...)  // tapered
let minted_since = (1.0 + inflation_now / nominal_epochs_per_year).powi(...)           // flat, today's rate
```

The function establishes in its own first branch that the past rate was higher than `inflation_now`,
then un-mints the supply as if every one of those epochs ran at `inflation_now`. Minting is
understated, so the reconstructed past supply is too large — and `supply` feeds
`InflationApyCalculator::estimate_yields` directly as `supply * inflation`.

Negligible at `epochs_behind` 0-1, which is the normal path. But `a_deep_backfill_cannot_exceed_the_rate_the_curve_started_at`
exercises 2000 epochs back, so deep backfill is a considered use case; there the average true rate is
~6.5% against the 4.3% used, roughly a 25% supply overstatement. Either compound the tapered rate
epoch by epoch, or state in a comment that the supply figure is only valid for shallow backfills and
cap `epochs_behind` accordingly.

collect/src/validators_block_rewards.rs
```diff
@@ -175,10 +175,14 @@ pub fn collect_validator_block_rewards_info(
                 rewards_params.loading_limit,
             );
         } else {
-            let hours_since_epoch_end = current_epoch_info
-                .slot_index
-                .saturating_mul(MILLISECONDS_PER_SLOT)
-                / MILLISECONDS_PER_HOUR;
+            // The target epoch ended when the current one started.
+            let hours_since_epoch_end = seconds_since_epoch_start(&client, &current_epoch_info)
```
ilmari-h: ### 🤖 The "stakes-etl overdue" alarm now silently disables itself on any RPC failure** —
`collect/src/validators_block_rewards.rs:179-186`

```rust
.unwrap_or_else(|err| { warn!("Cannot measure time since the epoch started: {err}"); 0 })
```

Before this PR the value was pure arithmetic on an already-fetched `EpochInfo` and could not fail.
Now three RPC calls sit behind it, and any of them failing yields `0`, which is below every
`max_data_delay_hours` threshold — the escalation path can never trigger. The leniency is deliberate
and defensible, but it needs to be observable: a `warn!` inside a job that then reports success is
easy to miss. Suggest a metric, or at minimum stop printing the fabricated `~0h` in the success-path
message at line 197 — say `unknown` there instead, so the log does not assert something false.

check/src/validators_jito.rs
```diff
@@ -80,11 +80,18 @@ pub async fn check_jito(
...
             {
-                info!(
-                    "To execute required to wait at epoch {current_epoch} for slot index {}, approximately {} seconds",
-                    sql_slot_index + params.execution_interval_slots,
-                    (sql_slot_index + params.execution_interval_slots - current_slot_index) * Decimal::from(MILLISECONDS_PER_SLOT) / Decimal::from(1000)
-                );
+                let target_slot_index = sql_slot_index + params.execution_interval_slots;
+                let slots_to_wait = target_slot_index - current_slot_index;
+                // An unavailable ETA must not turn "not yet" into a failed check.
+                match measure_milliseconds_per_slot(rpc_client, &epoch_data).unwrap_or_default() {
```
ilmari-h: ### 🤖 Three RPC round trips per check tick to format one log line** —
`check/src/validators_jito.rs:86`, `check/src/validators_block_rewards.rs:73-74`

`measure_milliseconds_per_slot` is called purely for the ETA text. `.unwrap_or_default()` on
`Result<Option<u64>>` also swallows the error with no log at all, so a persistently failing RPC path
here is invisible. Log the error at debug level, and consider whether the ETA is worth the calls.

collect/src/common.rs
```diff
@@ -34,10 +38,127 @@ where
...
+    let first_slot = epoch_info
+        .absolute_slot
+        .saturating_sub(epoch_info.slot_index);
+    let anchor = first_slot.max(rpc_client.get_first_available_block()?);
+    // The anchor slot may be skipped, and `get_block_time` only answers for produced blocks.
+    let anchor_block = *rpc_client
+        .get_blocks_with_limit(anchor, 1)?
+        .first()
+        .context("No block produced yet since the epoch start")?;
+    let elapsed = chrono::Utc::now().timestamp() - rpc_client.get_block_time(anchor_block)?;
```
ilmari-h, line 60: ### 🤖 The measurement window is floored in slots, not in seconds** — `collect/src/common.rs:8,57`

`MIN_SLOTS_TO_MEASURE = 1000` is 400 s at the baseline but only 200 s at 200ms. `elapsed` is
`chrono::Utc::now()` (local container clock) minus `get_block_time` (the cluster's stake-weighted
estimate). Both sides can be tens of seconds off; over a 200 s window that is a >10% rate error.

Consequence is not just a false "Check the slot time gate pubkeys" warning. In the pruned-ledger
branch (`common.rs:63-70`) the biased rate is extrapolated across the whole epoch via
`seconds_from_rate`, and that number is what `max_data_delay_hours` compares against — a 15% high
reading turns 12 h into ~13.8 h and hard-fails the block rewards job early. Block time typically lags
wall clock, so the bias has a direction: toward premature failure.

Add a wall-clock floor alongside the slot floor (e.g. also require `elapsed_seconds >= 600`).

ilmari-h (commented):
Read through the changes and they make sense to me. Adding a few findings from Claude that can hopefully be helpful.

collect/src/validators_performance.rs
ochaloup: 🤖 claude: Fixed — `warn_on_slot_time_divergence` now runs right after `get_slot_params`, outside the `with_rewards` arm, so every run that persists `slots_per_year` also cross-checks it. The extra RPC calls on a `--with-rewards`-off run are noise next to the vote-account and block-production fetches already in that path.

collect/src/validators_performance.rs
ochaloup: 🤖 claude: Fixed — the walk now compounds epoch by epoch, un-minting at each epoch's own tapered rate instead of today's flat one, which drops the reconstructed supply 28% at 2000 epochs back while leaving the shallow one-epoch path bit-identical. Pinned by a new `a_deep_backfill_un_mints_at_the_rates_those_epochs_ran` test.

collect/src/common.rs
ochaloup: 🤖 claude: Fixed — added `MIN_SECONDS_TO_MEASURE = 600` alongside the slot floor. Worth flagging that this deliberately rejects 1000 slots at the 400ms baseline too (400s), so measurement now starts ~10 minutes into an epoch and the old `milliseconds_per_slot(400, MIN_SLOTS_TO_MEASURE).is_some()` assertion is inverted.

collect/src/validators_block_rewards.rs
ochaloup: 🤖 claude: Fixed the log — `hours_since_epoch_end` is an `Option` now and both messages print `unknown` rather than a fabricated `~0h`, with `None` still never escalating. Declining the metric: this crate has no metrics channel today and adding one for a single gauge belongs outside this PR.

check/src/validators_jito.rs
ochaloup: 🤖 claude: Fixed the visibility half — both sites now `unwrap_or_else` with a `debug!`, so a persistently failing RPC path leaves a trace. Keeping the ETA itself: the three calls are skipped outright for the first 1000 slots of an epoch, and dropping it would regress the very log quality this PR set out to fix.

collect/src/slot_params.rs
ochaloup: 🤖 claude: Checked the transcription against agave `runtime/src/slot_params.rs` and `feature-set/src/lib.rs` on master — all four gate pubkeys and all five `slots_per_year` rows match byte for byte, so the hand-copied table is correct as it stands. Leaving it; it can be swapped for the crate import whenever the workspace solana deps get bumped.

ochaloup:
🤖 claude: Declining the `SLOTS_IN_EPOCH` → `EpochSchedule.slots_per_epoch` nitpick — the constant is also interpolated into the `/rewards` SQL in `store/src/rewards.rs` and used by `InflationApyCalculator`, so making only `collect` dynamic would desync the two. Doing it properly means storing `slots_per_epoch` per epoch, which is its own change.

Separately, the `epoch > activation_epoch` rule in `select_slot_params` is now confirmed against agave rather than assumed: `SlotParamsArchive::feature_effective_slot` maps an activation in epoch E to the first slot of E+1.

collect/src/validators_performance.rs
```diff
@@ -97,11 +96,95 @@ pub struct ValidatorsPerformanceSnapshot {
...
+    })?;
+
+    let (inflation, sol_total_supply) = inflation_and_supply_at(
+        rate.total,
+        client.supply()?.value.total,
+        governor.taper,
+        governor.initial,
+        governor.terminal,
+        slots_per_year / SLOTS_IN_EPOCH as f64,
+        epochs_behind,
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — at `epochs_behind == 1` the single un-mint step covers the target epoch itself, so that epoch's own `slots_per_year` is the exactly correct value to use, and `epochs_behind == 0` returns early; the mismatch needs a span of 2+ epochs crossing an activation. That span is not reachable here: for a non-current epoch `get_block_production_by_validator` falls back to `get_leader_schedule_with_commitment(...).unwrap()`, which a node answers with null for a historical epoch, so the snapshot fails loudly before any inflation figure is written.

ochaloup:
@ilmari-h thank you for the review. I addressed most of it or added an explanation of what and why not. May I ask for a re-review?

ilmari-h (dismissed):

ilmari-h (approved):

PR #253, opened 2026-08-17
[GEN-8006][collect] reuse the fetched vote accounts for credits

ochaloup:
get_credits re-fetched getVoteAccounts (~300 KB) although its caller already held the same response, and only ever reads epoch_credits from it. The per-minute performance collector paid for that twice a minute.

Taking the data instead of the client also makes credits, activated stake, commission and delinquency come from one coherent read rather than two fetches that could straddle a slot boundary.

## Summary by CodeRabbit

* **Performance**
  * Improved validator performance data collection by reusing available vote-account information.
  * Reduced redundant network requests during epoch credit processing.
* **Reliability**
  * Simplified credit retrieval and reduced errors related to additional RPC requests.

### DIFF
--- a/collect/src/solana_service.rs
+++ b/collect/src/solana_service.rs
@@ -60,10 +60,8 @@ pub fn get_stake_history(rpc_client: &RpcClient) -> anyhow::Result<StakeHistory>
     )?)
 }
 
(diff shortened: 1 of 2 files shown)


ilmari-h (approved):
nice, less RPC calls always good

PR #254, opened 2026-08-17
[GEN-8714][api] take rate adjustments + ASN query fix

ochaloup:
Stops `expected_take_rate` from reading at the honest floor for validators that advertise a low commission early in the epoch and raise it before rewards are assessed.

- Project `commission_max_observed`, `commission_min_observed` and `commission_effective` from the newest closed epoch, since seeding the record from the still-open epoch left all three null for every validator
- Feed `expected_take_rate` the worse of the observed ceiling and the advertised rate, so a fee rise counts at once while a cut waits for the epoch to close
- Route `order_field=Commission` through the same selector, which until now tied every validator at 100 and made that ordering meaningless
- Keep a validator's stored data centre when a whois lookup fails, but only on an unchanged IP and only when nothing resolved, so a partial answer never pairs a new network with the previous city
- Pin both fixes with database-backed tests through the real `load_validators` and the validator upsert, and share the snapshot fixture the three store tests were each copying

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validator eligibility checks using recent epoch activity, stake, and credits.
  * Commission rankings and projected take rates now use the highest known commission, reducing misleading estimates.
  * Preserved validator data-center information when lookups are temporarily unresolved or data is missing.
  * Correctly refreshes or clears data-center details when validator locations change.

* **Enhancements**
  * Added support for serializable validator grouping and hierarchy data, including stake, APY, delegation, and epoch details.
  * Expanded coverage for commission calculations and data-center persistence across historical records.


store/src/validators.rs
```diff
@@ -65,14 +65,15 @@ pub async fn store_validators(
...
+            dc_coordinates_lon = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_coordinates_lon ELSE validators.dc_coordinates_lon END,
+            dc_continent = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_continent ELSE validators.dc_continent END,
+            dc_country_iso = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_country_iso ELSE validators.dc_country_iso END,
+            dc_country = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_country ELSE validators.dc_country END,
+            dc_city = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_city ELSE validators.dc_city END,
+            dc_asn = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_asn ELSE validators.dc_asn END,
+            dc_aso = CASE WHEN u.dc_resolved OR u.node_ip IS DISTINCT FROM validators.node_ip THEN u.dc_aso ELSE validators.dc_aso END,
```
ilmari-h, line 78: ### 🤖 Data-center guard misses the epoch boundary

The guard sits only in the UPDATE branch, which matches on `validators.epoch = u.epoch`. The first store of a new epoch has no row for that epoch, so it takes the INSERT branch (`store/src/validators.rs:239`) and writes `v.dc_*` raw. If whois is down at that moment the new row gets eight nulls with no carry-over from the previous epoch — and every later unresolved run in the epoch now *keeps* them, because the guard preserves what is stored.

`load_validators` seeds `dc_*` from the newest epoch row, so the validator reads `Unknown` city/ASN/ASO for the rest of the epoch and lands in the `Unknown` bucket of `load_dc_concentration_stats`. Same symptom the PR fixes, one write earlier.

**Suggested fix:** when `dc_resolved` is false, backfill `dc_*` on insert from that validator's most recent earlier epoch row with a matching `node_ip` — either a subquery in the INSERT or a follow-up UPDATE over the freshly inserted vote accounts.

store/src/utils.rs
```diff
@@ -1188,6 +1196,15 @@ pub async fn load_validators(
                     rugged_commission_occurrences: 0,
                 });
 
+            // Gating all three on max_observed keeps them from one row, the newest closed epoch; commission_advertised deliberately stays on the open epoch that seeded the record.
+            if record.commission_max_observed.is_none() {
+                record.commission_max_observed =
+                    row.get::<_, Option<i32>>("commission_max_observed");
+                record.commission_min_observed =
+                    row.get::<_, Option<i32>>("commission_min_observed");
+                record.commission_effective = row.get::<_, Option<i32>>("commission_effective");
```
ilmari-h: ### 🤖 Commission projection can reach back past the newest closed epoch

`store/src/utils.rs:1200` — rows arrive `ORDER BY epoch DESC`, so the `is_none()` gate keeps reassigning all three columns from each successively older row until one has a non-null `commission_max_observed`. If the newest closed epoch never got it written (`close_epoch` not run for it yet, or the rewards row is missing), the record publishes commission from an arbitrarily older epoch — up to `display_epochs` back — as current, and both `expected_take_rate` and `order_field=Commission` consume it with no marker that it is stale. If no row in the window has it, the three end up from the *oldest* row rather than staying null.

The values always come from one row, so there is no cross-column epoch mixing; only *which* row is wrong.

**Suggested fix:** pin to the single newest closed epoch — read the row's `epoch` alongside the values and accept only the first row with `epoch` below the record's seeding epoch, leaving the three null when that row has them null.

store/src/validators.rs
ochaloup: 🤖 claude: Fixed — the INSERT path now carries the previous epoch's `dc_*` when the lookup came back unresolved and `node_ip` is unchanged, reaching back at most 10 epochs and skipping epochs that hold no location at all. Four tests cover it (carry, IP change, partial answer, stepping over an empty epoch), each proven by reverting the guard and watching exactly that test fail.

store/src/utils.rs
ochaloup: 🤖 claude: Fixed — the walk now stops one epoch below the record's own seeding epoch, so the three columns come from the newest closed epoch or stay null, and `worst_known_commission` falls back to the fresh `commission_advertised` so the take rate never blanks. One correction: a missing rewards row cannot null `commission_max_observed` by itself, since `GREATEST` ignores NULL arguments and `commission_advertised` is always set — the reachable trigger is `close_epoch` not having run for that epoch.

store/src/validators.rs
```diff
@@ -335,10 +341,67 @@ pub async fn store_validators(
...
+                unresolved_insertions.push(vote_account.clone());
+            }
         }
         insertions += query.execute(psql_client).await?.unwrap_or(0);
         info!("Stored {insertions} new validator records");
     }
 
+    if !unresolved_insertions.is_empty() {
+        carry_previous_data_centers(psql_client, &snapshot_epoch, &unresolved_insertions).await?;
+    }
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — unresolved accounts are now collected in the UPDATE branch too, and the carry target is restricted to rows where `num_nonnulls(dc_*) = 0` so a preserved or resolved location is never overwritten. Two tests cover it: one proves a row left empty by an earlier run gets repaired later in the same epoch, the other proves the epoch's own location survives the repair — each fails when exactly its half of the fix is reverted.

deny.toml
```diff
@@ -44,6 +44,14 @@ ignore = [
...
   # Revisit when gcp-bigquery-client moves to a rustls 0.23 release.
   "RUSTSEC-2026-0049",
+  # h2 unbounded empty DATA frames — queued without limit, so an unread stream
+  # can grow memory or panic on overflow. Low severity, DoS only. The advisory
+  # patches ">= 0.4.16" and nothing in the 0.3 line: 0.4.x is updated to 0.4.16
+  # here, but h2 0.3.26 stays, reached through hyper 0.14 from warp 0.3 (the API
+  # server, the one side that actually accepts frames) and from reqwest 0.11 /
+  # solana-client 2.2. Getting off it means a hyper 1.x server for warp and
+  # reqwest 0.11 -> 0.12, which waits on the Solana client crates.
+  "RUSTSEC-2026-0258",
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining the waiver format — every ignore in this file is a bare id with a prose comment, and an expiry firing inside the org-wide static-analysis workflow would break unrelated builds on a date nobody watches. The reachable point was right though, so the comment now separates the two ends: the client side waits on the Solana crates, while closing the exposed warp/hyper listener is repo-local work that waits on nobody.

ilmari-h (dismissed):

ochaloup:
@ilmari-h updated to fix the merge conflicts. :pray: re-review

ilmari-h (approved):

PR #256, opened 2026-08-18
[GEN-8733] [GEN-8734] endpoints to list validator clients and providers and their aggregate stats

ilmari-h:
Adds 2 new endpoints

Both support:
- Sorting by: Available values : Name, Stake, StakeDelta7d, StakeDelta30d, NetApy, TakeRate, Validators, Delegators
- Text search filtering
- pagination with `offset` and `limit` just like the existing `/validators` endpoint

### `/clients`

Returns list of Solana clients and their aggregate stats. Each client entry also carries a breakdown of block engines and their aggregate stats.

Example response:

```
{
  "clients": [
    {
      "group": {
        "key": "Agave",
        "validator_count": 623,
        "total_stake": "372043980923007769",
        "stake_share": 0.853944907115434,
        "stake_delta_7d": "750175454321447",
        "stake_delta_30d": null,
        "net_apy": 0.0382142971976767,
        "take_rate": 0.358233883161141,
        "delegator_count": 628492
      },
      "children": [
        {
          "key": "Agave + JitoBAM",
          "validator_count": 375,
          "total_stake": "143012052576089898",
          "stake_share": 0.328252626612845,
          "stake_delta_7d": "1283744715230488",
          "stake_delta_30d": null,
          "net_apy": 0.0383160766962081,
          "take_rate": 0.356457803127274,
          "delegator_count": 315708
        },
        {
          "key": "Agave + Jito",
          "validator_count": 153,
          "total_stake": "121066067268127138",
          "stake_share": 0.277880457336323,
          "stake_delta_7d": "29359060030389343",
          "stake_delta_30d": null,
          "net_apy": 0.0435434359348923,
          "take_rate": 0.259541422947144,
          "delegator_count": 269822
        },
        ...
    }
}
```

### `/providers`

Returns list of providers and their validator aggregate stats.

Example response

```
{
  "providers": [
    {
      "key": "TeraSwitch Networks Inc.",
      "validator_count": 75,
      "total_stake": "77341899568047753",
      "stake_share": 0.177521273369117,
      "stake_delta_7d": "-58616557570685826",
      "stake_delta_30d": "-37202680536283215",
      "net_apy": 0.034608558612931,
      "take_rate": 0.427316928292815,
      "delegator_count": 181866
    },
    {
      "key": "UAB Cherry Servers",
      "validator_count": 83,
      "total_stake": "58221130029393997",
      "stake_share": 0.133633763814056,
      "stake_delta_7d": "3087309436209535",
      "stake_delta_30d": "4302598415688444",
      "net_apy": 0.0271923246957011,
      "take_rate": 0.545508468421133,
      "delegator_count": 44029
    },
    {
      "key": "Latitude.sh",
      "validator_count": 99,
      "total_stake": "44567420498050699",
      "stake_share": 0.102294684792813,
      "stake_delta_7d": "5517211166218834",
      "stake_delta_30d": "7692622824935945",
      "net_apy": 0.0456951312466505,
      "take_rate": 0.226605192955891,
      "delegator_count": 75090
    },
    ...
}
```

## Summary by CodeRabbit

- **New Features**
  - Added client and provider listings with search, sorting, pagination, stake totals, validator counts, and APY metadata.
  - Added validator grouping by client, lineage, and hosting provider.
  - Added hierarchical client views and configurable group ordering.
  - Added OpenAPI documentation for the new endpoints and response formats.

- **Bug Fixes**
  - Improved validator eligibility handling using recent statistics and the latest reported epoch.

- **Performance**
  - Precomputed and cached client and provider group summaries for faster responses.


deny.toml
```diff
@@ -163,7 +169,11 @@ wildcards = "deny"
 # otherwise flag as a wildcard. Allow wildcards *only* for path deps; a
 # wildcard on a crates.io dependency is still denied.
 allow-wildcard-paths = true
-deny = []
+deny = [
+  # The advisories section ignores RUSTSEC-2026-0258 for the unfixable h2 0.3
+  # line; the 0.4 line is patched in 0.4.16, so keep it from sliding back.
+  { crate = "h2:>=0.4, <0.4.16", reason = "RUSTSEC-2026-0258, patched in 0.4.16" },
```
ochaloup, line 175: Nice that the static checks are clean 🤟

store/src/groups.rs
```diff
@@ -0,0 +1,1052 @@
...
+        .filter(|(_, epoch_end_at)| **epoch_end_at <= cutoff)
+        .map(|(epoch, _)| *epoch)
+        .max()
+}
+
+type ReferenceStake = HashMap<FoldedKey, Decimal>;
+
+/// Stake per group as it stood in `epoch`, bucketed by that epoch's own client and provider values.
+/// `None` when that epoch classified nothing; client ids reach back only to epoch 1011.
+fn stake_by_key_at(
```
ochaloup, line 225: 🤖 claude: `stake_by_key_at` receives the already-`eligible()`-filtered slice, so the 7d/30d baselines exclude validators that have since dropped out of the eligible set — a provider that lost 3M SOL because three validators stopped voting reports a flat or positive `stake_delta_7d` instead of -3M. Suggest computing the baselines over the full validator map before `eligible()` narrows it, and fixing the doc comment, which claims a baseline this does not compute.

store/src/groups.rs
```diff
@@ -0,0 +1,1052 @@
...
+            self.net_apy_weighted += net_apy * weight;
+            self.net_apy_weight += weight;
+        }
+
+        if let Some(take_rate) = validator.avg_take_rate.filter(|rate| rate.is_finite()) {
+            self.take_rate_weighted += take_rate * weight;
+            self.take_rate_weight += weight;
+        }
+
+        if let Some(unique_delegators) = validator.unique_delegators {
```
ochaloup, line 136: 🤖 claude: `unique_delegators` is `COUNT(DISTINCT withdraw_authority)` per vote account, so summing it across a group counts one authority once per validator it stakes to — Marinade's own stake authority alone inflates the `Agave` row by roughly the Agave validator count. A true per-group distinct count needs a new group-keyed BQ query, so the cheap fix is to rename/document the field as delegation relationships rather than delegators.

store/src/groups.rs
```diff
@@ -0,0 +1,1052 @@
...
+        }
+    }
+
+    stake_by_key
+        .keys()
+        .any(|key| key.is_some())
+        .then_some(stake_by_key)
+}
+
+pub fn aggregate_groups(
```
ochaloup: 🤖 claude: `aggregate_groups` has no non-test callers (production reaches only `aggregate_all`), and `GroupKind` plus `DELTA_SHORT_DAYS`/`DELTA_LONG_DAYS` are `pub` with no reader outside this module. Suggest `#[cfg(test)]` on `aggregate_groups` and dropping the other `pub`s; `UNKNOWN_GROUP` should stay public since the api tests import it.

api/src/handlers/validator_groups.rs
```diff
@@ -0,0 +1,527 @@
...
+        };
+        ordering
+            .then_with(|| a.key.to_lowercase().cmp(&b.key.to_lowercase()))
+            .then_with(|| a.key.cmp(&b.key))
+    });
+
+    keyed.into_iter().map(|(_, group)| group).collect()
+}
+
+fn filter_groups(
```
ochaloup, line 111: 🤖 claude: `filter_groups` — the whole `/providers` query path — has no test: every `page_groups` test leaves `query: None`, and the `page_tree` query tests exercise `matches_query` instead. Worth covering `query=hetz` (rows filtered, `total_count` reflects the match) and a whitespace-only `query` (filter skipped); separately, 5 of 8 `GroupOrderField` extractors are unexercised, so swapping `StakeDelta7d` and `StakeDelta30d` would pass CI.

store/src/dto.rs
```diff
@@ -227,7 +227,7 @@ impl Validator {
     }
 }
 
-#[derive(Deserialize, Serialize, Debug, Clone, utoipa::ToSchema)]
+#[derive(Deserialize, Serialize, Debug, Clone, Default, utoipa::ToSchema)]
```
ochaloup: 🤖 claude: `Default` here is used only by the same-crate fixtures in `store/src/groups.rs`; production builds these as full struct literals, and `ValidatorRecord::default()` would serialize a blank `identity`/`vote_account`. `#[cfg_attr(test, derive(Default))]` on both this and `ValidatorRecord` keeps the fixtures working without widening the production API — `ValidatorGroupRecord`'s `Default` must stay unconditional for the api tests.

ochaloup (commented):
Few points from Claude. In general it seems all good to me.

store/src/groups.rs
ilmari-h: done

store/src/dto.rs
ilmari-h: fixed

api/src/handlers/validator_groups.rs
ilmari-h: added tests

api/src/handlers/list_clients.rs
```diff
@@ -0,0 +1,85 @@
...
+use warp::{http::StatusCode, reply::json, Reply};
+
+/// Clients as a two-level tree: one row per client, its block engines underneath.
+#[derive(Serialize, Debug, utoipa::ToSchema)]
+pub struct ResponseClients {
+    clients: Vec<ValidatorGroupNode>,
+    total_count: usize,
+    total_activated_stake: Decimal,
+    current_epoch: Option<u64>,
+    net_apy_updated_at: Option<DateTime<Utc>>,
```
ochaloup, line 21: 🤖 claude: `ResponseProviders` documents these same four fields and `utoipa` copies those doc comments into the OpenAPI schema, so `/clients` publishes the units and epoch semantics blank while `/providers` publishes them. Copy the four doc comments across.

api/src/handlers/list_clients.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let config = GetGroupsConfig {
+        order_field: query_params.order_field.unwrap_or(DEFAULT_ORDER_FIELD),
+        order_direction: query_params
+            .order_direction
+            .unwrap_or(DEFAULT_ORDER_DIRECTION),
+        offset: query_params.offset.unwrap_or(0),
+        limit: query_params.limit.unwrap_or(DEFAULT_LIMIT),
+        query: query_params.query,
+    };
```
ochaloup, line 62: 🤖 claude: A shared `GetGroupsConfig::from_params(..)` would be worth having, but not for CodeRabbit's stated reason — `DEFAULT_ORDER_FIELD`, `DEFAULT_ORDER_DIRECTION` and `DEFAULT_LIMIT` are already consts that both handlers import, so the two endpoints cannot silently diverge on a default. What is actually duplicated is eight lines of unwrapping, which makes this a plain DRY nit rather than a divergence risk.

api/src/handlers/validator_groups.rs
```diff
@@ -0,0 +1,630 @@
...
+            None => true,
+            Some(query) => matches_query(node, query),
+        })
+        .collect();
+    let total_count = matching.len();
+
+    let mut children_by_key: HashMap<String, Vec<ValidatorGroupRecord>> = matching
+        .iter()
+        .map(|node| (node.group.key.clone(), node.children.clone()))
+        .collect();
```
ochaloup, line 191: 🤖 claude: `children_by_key` clones the children of every matching node, including the ones `skip`/`take` then discards; a `for` loop moving `node.children` out of `matching` is the same length and clones nothing. Optional — the client tree is a few dozen small structs per request, so this is cleanliness rather than a hot path.

store/src/groups.rs
```diff
@@ -0,0 +1,1096 @@
...
+        .epochs
+        .delta_7d
+        .and_then(|epoch| stake_by_key_at(&population.all, epoch, kind));
+    let stake_long = population
+        .epochs
+        .delta_30d
+        .and_then(|epoch| stake_by_key_at(&population.all, epoch, kind));
+
+    let mut accumulators: HashMap<FoldedKey, Accumulator> = Default::default();
+    for validator in &population.eligible {
```
ochaloup, line 310: 🤖 claude: The baseline now reads `population.all` while the current row reads `population.eligible`, which is correct at steady state but wrong while an epoch is only partially written — `store_validators` inserts in chunks of 500 and `eligible()` takes `last_epoch` from the max across the whole map, so the first chunk of epoch N+1 makes every not-yet-written validator ineligible. If the 10-minute cache refresh lands in that window, every client and provider row serves a cluster-scale negative `stake_delta_7d`/`stake_delta_30d` until the next refresh; cheapest guard is to leave both as `None` when `eligible.len()` is a small fraction of `all.len()`.

ochaloup (commented):
A small follow-up on fixes. Talked with claude, feel free to decline, those seems not a big severity.

store/src/groups.rs
ilmari-h: we only deal in complete epochs as they're afaik

api/src/handlers/validator_groups.rs
ilmari-h: seems like nit pick

store/src/groups.rs
ilmari-h: turns out counting the unique delegators per group is not so simple so for now just renamed this to `delegation_relationships`. Will implement distinct counting in a separate PR.

ilmari-h:
@ochaloup thanks, indeed claude's findings seemed like nitpicks this time. Ready from my end

ochaloup (approved):

PR #259, opened 2026-08-20
[GEN-8772] delete the repo-local cargo-deny override

janlegner:
Deletes `deny.toml` (194 lines) and bumps `rustls-webpki 0.103.8 → 0.103.14`. One of its 22 ignores was masking a live, one-command-fixable vulnerability: the entry's own comment claimed repos pick up the fix via `cargo update`, but the lockfile had never moved, so RUSTSEC-2026-0049 was suppressed rather than fixed — that bump is the fix.

The rest: 3 dead, 10 dissolved by `unmaintained = "workspace"`, 3 org-covered, 4 irreducible (`h2 0.3.27` with no patched release, the dalek pair now accepted org-wide, `bincode`). The four `unlicensed` findings clear under the org policy's `private.ignore`, so no licence needs choosing.

Also migrates structopt to clap 4's derive feature across 24 files, which clears RUSTSEC-2022-0104 — the last self-fixable finding here. The substitution turned out to be purely mechanical: `StructOpt` → `Parser`, `#[structopt(...)]` → `#[arg(...)]` (or `#[command(...)]` for the 6 `flatten`/`subcommand` sites), `short = "u"` → `short = 'u'` at 4 sites, `from_args()` → `parse()`, and `from_iter()` → `parse_from()` in the two test helpers. There were no `parse(...)`, `arg_enum` or `AppSettings` attributes to translate, and clap 4's `Parser` derive generates `Args` and `Subcommand` too, so flattened structs and subcommand enums needed no separate derive.

The lockfile shrinks by 5 packages: structopt, structopt-derive, proc-macro-error, proc-macro-error-attr, ansi_term, strsim, textwrap, unicode-width and vec_map go, clap_builder, clap_derive, clap_lex and heck arrive, and clap moves 2.34.0 → 4.5.60.

`cargo deny --all-features --config .github/deny.toml check` goes from 2 errors to 1 against the live org policy; `cargo +1.88.0 check --workspace --all-targets`, `cargo clippy --workspace --all-targets` and `cargo fmt -- --check` are all clean with zero warnings. The lockfile stays at `version = 4` and the highest `rust-version` in the graph is unchanged at 1.88.0 (time 0.3.47), matching the toolchain pin — clap 4.5.60's own MSRV is 1.74.

The one remaining error is RUSTSEC-2026-0258 on h2 0.3.27, which has no fix on the 0.3 line (`patched = [">= 0.4.16"]`). It had three independent roots; two are now gone:

- `solana-client`/`solana-vote-program`/`solana-account-decoder` 2.2.20 → **2.3.13**. `solana-metrics` requires `reqwest ^0.11.27` through all of 2.2.x and moves to `^0.12.16` at 2.3.0, so the whole 2.3 line clears it. This forced `solana-config-program` → `solana-config-program-client 0.0.2`: config-program has no 2.3 release and pins `solana-program-runtime =2.2.20`, which cannot co-resolve with vote-program 2.3's `=2.3.13`. It is the same crate `solana-account-decoder` 2.3 uses; `id()` → `::ID`, same `Config111…` address.
- Workspace `reqwest` 0.11.27 → **0.12.16**, one source fix: `.status(status)` → `.status(status.as_u16())` in `api/src/handlers/reports_scoring_html.rs`, since reqwest 0.12 is http 1.x and warp 0.3 is http 0.2. `chrono`'s `serde` feature and `reqwest`'s `blocking`/`json` are now declared directly — they had been arriving only through feature unification from `solana-config-program` and `solana-metrics`, and chrono's was load-bearing.

**The remaining root is `warp 0.3.7` → hyper 0.14, and it is a service-behaviour decision, not a bump.** warp 0.4 is otherwise mechanical (`default = []`, so it needs `features = ["server","multipart"]`; multipart, reply, cors, serve and filters are unchanged), but its response body is a private type constructible only `From<Bytes|String|Vec<u8>|&'static str>`. There is no public equivalent of `hyper::Body::wrap_stream`, so `reports_scoring_html.rs:51` would have to buffer the entire proxied report instead of streaming it. Left for the repo owners; moving that handler to axum 0.8, which does have streaming bodies, is the alternative worth weighing.

`cargo build --release --locked` and clippy both exit 0, and the test suite passes — 8 binaries, 35 tests.

⚠️ Merge after marinade-finance/.github#6 — cargo-deny is red on the `jito-programs` git source, which only that policy's `[sources.allow-git]` clears.

---
_Generated by [Claude Code](https://claude.ai/code)_

## Summary by CodeRabbit

* **New Features**
  * Updated command-line interfaces with improved argument parsing and environment-variable support.
  * Added multipart request handling for API interactions.
  * Upgraded Solana, HTTP, and date/time integrations for improved compatibility.

* **Bug Fixes**
  * Improved API report responses while preserving appropriate status codes, content types, and error messages.
  * Updated validator configuration lookups to use the current client integration.


janlegner:
Final CI state on `98014dd` — 10 checks green, 1 red:

| Check | Result |
|---|---|
| `lint-and-test` | ✅ success |
| `static-analysis / clippy` | ✅ success |
| `static-analysis / detect` | ✅ success |
| `static-analysis / cargo-deny` | ❌ **failure** |
| `static-analysis / solana-lints` | ⏭️ skipped |
| `static-analysis / sec3-xray` | ⏭️ skipped |
| `CodeQL` | ✅ success |
| `Analyze (actions)` | ✅ success |
| `Trivy Scan` | ✅ success |
| `Scan` | ✅ success |
| `check-dependabot` | ✅ success |

`lint-and-test` passing (`cargo fmt --check && cargo clippy`, then `cargo test --all-features`) confirms the `rustls-webpki` 0.103.8 → 0.103.14 lockfile bump is safe.

`static-analysis / cargo-deny` reports exactly the 6 findings predicted in the description — `advisories FAILED, bans ok, licenses ok, sources FAILED`:

- 3 × `source-not-allowed` — `jito-tip-distribution`, `jito-priority-fee-distribution`, `jito-programs-vote-state` from `jito-foundation/jito-programs`. **Already fixed by the proposed org policy's `[sources.allow-git]`; needs nothing from this repo.**
- RUSTSEC-2026-0258 — `h2` 0.3.27, no patched 0.3 release.
- RUSTSEC-2022-0104 — `structopt`, direct dep, needs the clap 4 migration.
- RUSTSEC-2024-0320 — `yaml-rust` via `serde_yaml` 0.8. **Fixed by the proposed policy's `unmaintained = "workspace"`** (it is transitive), so this one also disappears when the org policy lands.

So of the 6, four go away with the org policy alone. The residual pair after it lands is `h2` 0.3.27 (upstream) and `structopt` (in-repo migration), plus `bincode` and the two dalek crates and the 4 missing `license` fields that the proposed policy newly surfaces — all itemised in the description.

The org config's own RUSTSEC-2026-0049 entry is now reported as `advisory-not-detected` on this branch, confirming the `rustls-webpki` bump resolved that advisory rather than suppressing it.


---
_Generated by [Claude Code](https://claude.ai/code)_

ilmari-h (approved):

PR #261, opened 2026-08-21
[GEN-8735] validator grouping by operator

ilmari-h:
This PR adds a new parameter to endpoint `GET /validators`: `with_operator_groups`.

When set to `true` the response will include property `operators`. 

Example:
```json
 "operators": [
    {
      "name": "Coinbase",
      "validator_count": 25,
      "total_stake": "40033932372918425",
      "stake_share": 0.0918890626233035,
      "stake_delta_7d": "36023751220531",
      "stake_delta_30d": "328681958281758",
      "net_apy": 0.0203416949988198,
      "take_rate": 0.672546639762482,
      "credits": 3630677.049966,
      "marinade_score": null,
      "apy": 0,
      "commission": 62.4522682438837,
      "uptime_pct": 16.5522195505889,
      "expected_take_rate": 0.664166095396705,
      "delegation_relationship_count": 4689,
      "incident_count_3m": 0
    },
    {
      "name": "Helius",
      "validator_count": 1,
      "total_stake": "16003006030071666",
      "stake_share": 0.0367313710169801,
      "stake_delta_7d": "33961831028939",
      "stake_delta_30d": "97098232056798",
      "net_apy": 0.0538651086824825,
      "take_rate": 0.082171097,
      "credits": 3630731,
      "marinade_score": null,
      "apy": 0,
      "commission": 0,
      "uptime_pct": 16.552056427067,
      "expected_take_rate": 0.105581378964153,
      "delegation_relationship_count": 13946,
      "incident_count_3m": 0
    }
    // and many more
  ]
```
Each validator also gains `operator` property, that can be used to map it to the entry in `operators`:

```json
 "validators": [
    {
      "identity": "Fd7btgySsrjuo25CJCj7oE7VPMyezDhnx7pZkj2v69Nk",
      "vote_account": "CcaHc2L43ZWjwCHART3oZoJvHLAe9hzT2DJNUpBzoTN1",
      "operator": "Coinbase",
      // etc...
    },
    // etc...
  ]
```

When `with_operator_groups` is used, the returned `validators` list is also sorted so that it follows the order of operators first, and any sorting with `order_field` applies after. `operators` sorting is dictated by query parameter `order_field`. Validators that don't belong to any group are treated as their own singleton groups, and can be sorted after or between the operator groups, instead of e.g. tailing them.

When `with_operator_groups` is enabled, pagination applies on the "row" level. E.g. one operator group is one row. So when you ask for a page of size 2 you might get this kind of response where there are more validators than the page limit:

```json
 "validators": [
    // 1st row
    {
      "identity": "A",
      "operator": "Coinbase"
    },
    {
      "identity": "B",
      "operator": "Coinbase"
    },
    {
      "identity": "C",
      "operator": "Coinbase"
    },

    // 2nd row
    {
      "identity": "D",
      "operator": "Binance"
    }
  ]
```

## New Features
- Added operator information and grouping for validator results.
- Added seven-day and 30-day stake-change metrics.
- Added shared sorting options across validators, clients, providers, and operator groups.
- Added sorting by APY, uptime, incidents, commission, credits, expected take rate, and other metrics.
- Added complete operator-group data alongside validator results.

## Bug Fixes
- Improved grouped pagination to return complete operator blocks.
- Improved grouped sorting with deterministic name ordering.
- Kept unmapped validators in overall results while excluding empty operator groups.
- Improved handling of missing validator history in stake-change calculations.


honzsik (commented):
Reviewed from the `marinade-web` side, against the Explore design for grouped validators (operator rows with their validators nested, collapsible, standalone validators as plain top-level rows). The aggregation itself looks right — these are the places where the contract and that design don't meet yet.

### What already lines up

- `ValidatorGroupRecord` covers every column a group row renders: `name`, `validator_count`, `total_stake`, `stake_share`, `stake_delta_7d/30d`, `net_apy`, `take_rate`, `expected_take_rate`, `delegation_relationship_count`, `incident_count_3m`. `expected_take_rate` matters in particular — the Take rate column renders that field, not `avg_take_rate`, so the group row can now agree with its children.
- `ValidatorRecord.operator` is enough to map a child to its parent, and children sorting inside their operator block with the direction turning blocks and contents together is exactly the intended behaviour.
- Dropping unclassified validators from the operator rows while keeping them in the `stake_share` denominator is the right call here: the design has no "Unknown" operator row, and operator shares should still be shares of the cluster.
- `OrderField::Incidents` and `OrderField::DelegationRelationships` make the Delegators and Incidents columns sortable — they are unsortable plain headers in the web app today, so this unlocks two headers for us.
- An operator with `validator_count: 1` (Helius in your example payload) can be rendered as a plain validator row, which is what the design does.

### Blocking for the design

**1. Validators with no operator sort behind every operator.** `sort_validators_ranked` ranks an unmapped validator `usize::MAX`, so it lands after all operator blocks in both directions. `operators.csv` maps 54 vote accounts across 12 operators, so the rest of the network — several hundred validators, including large ones — queues up after the mapped 54.

The design interleaves top-level rows by the sorted column regardless of whether they are a group: the 4th row is a standalone validator with 12.48M SOL (Jupiter, not in `operators.csv`), sitting directly among operator rows. For that to work, an unmapped validator has to rank as a singleton group on its own column value rather than being pushed to the tail. Without it the grouped list is simply not stake-ordered, and no client-side fix exists while the list is server-paged.

**2. `offset`/`limit` cut validators, but the page is a list of top-level rows.** A page in the design is "N top-level rows", some expanded, some collapsed. Paging by validator means a group can straddle a page boundary (its header repeats on both pages, each time claiming the full `validator_count`), and a collapsed group still spends its children's slots, so the number of visible rows per page varies with what the user has collapsed.

`page_tree` already solves this shape for clients — page the parents, ship each parent's children whole. Could `with_operator_groups` page the same way? Otherwise we have to pull the whole list (`limit=2000`) and page client-side, and `total_count` stops being usable for the pager.

**3. No per-validator `stake_delta_7d` / `stake_delta_30d`.** The design shows the two Δ columns on child rows as well as on group rows. Group rows have the deltas; `ValidatorRecord` does not. We can only derive them from `epoch_stats`, which means re-requesting the array we currently suppress with `epochs: 0` *and* re-implementing your timestamp→epoch window selection — and if we get that selection even slightly different, children stop reconciling with their parent. Two fields on `ValidatorRecord` computed the same way as the group ones would settle it.

Related: `StakeDelta7d`/`StakeDelta30d` are group-only extractors, so clicking a Δ header sorts the operator rows and leaves the validators inside each block on the `vote_account` tiebreak. If per-validator deltas land, it'd be good for those two fields to order validators too. (Separately, Δ ordering is on absolute lamports while the column shows a percentage of where the row started — a known caveat on the Providers/Clients tables, but inside a group it becomes more visible.)

### Numbers that won't reconcile on screen

**4. `incident_count_3m` and the count we render use different rules.** Yours counts every incident with `start_at` inside 90 days, unfiltered by duration. The web app's per-row count keeps only incidents of at least 3 minutes of downtime (sub-3-min restart noise is not a reportable outage, and the same floor drives the validator detail page and your `query_incident_free` filter), but applies no date window — it counts the whole `incidents[]` array, which `load_incidents` fills over 90 *epochs*, roughly twice the window. So a group total won't equal the sum of the counts shown on its children in either direction. In the design, the Figment group shows 11 against children of 4 + 5 + 2.

Cleanest fix from our side: apply the 3-minute floor to `incident_count_3m`, and expose the per-validator value with the same rule so both sides read one number. That would also make our "Incidents (3mo)" header honest, which it currently isn't.

**5. Group aggregates ignore `query` and the status filters.** The doc comment is explicit that `operators` is never filtered, which is a defensible choice — but with a filter chip or a search term active, a group row states whole-group stake and validator count while only the matching children render underneath it. Is filter-aware aggregation feasible, or should we show group rows only in the unfiltered view? Happy either way, we just need to know which.

### One coordination note

`ValidatorGroupRecord.key` → `name` and folding `GroupOrderField` into `OrderField` are both breaking for consumers. Our types are generated from the live `docs.json`, and the Providers/Clients tabs that read `key` are on an unmerged branch, so nothing in production breaks *if* this deploys first. If the order flips, those tabs read `undefined` for every group name at runtime. API-first sequencing is fine for us — just flagging so it isn't landed the other way around.

store/src/dto.rs
```diff
@@ -550,22 +556,26 @@ pub struct FeatureSetStats {
...
 }
 
-/// One hosting provider or one validator client, aggregated over every validator running it in the
-/// last epoch.
+/// Group can be a hosting provider, validator client or node operator
 #[derive(Deserialize, Serialize, Debug, Clone, Default, utoipa::ToSchema)]
 pub struct ValidatorGroupRecord {
     /// Name of the group or `Unknown`
-    pub key: String,
+    pub name: String,
```
ochaloup: 🤖 claude: This renames the JSON property on `/clients` and `/providers` too — including the nested `children` of the client tree — not only on the new operator rows. @honzsik  confirmed marinade-web is fine as long as this deploys before their Providers/Clients branch, but is every other consumer of those two endpoints in sync on `key` → `name`?

api/src/handlers/list_validators.rs
```diff
@@ -240,6 +294,15 @@ fn get_field_extractor(order_field: OrderField) -> FieldExtractor {
...
         },
+        OrderField::DelegationRelationships => {
+            |a: &ValidatorRecord| a.unique_delegators.map(Decimal::from)
+        }
+        OrderField::Incidents => |a: &ValidatorRecord| Some(Decimal::from(a.incident_count_3m)),
+        // Group-only columns: they order the operator rows, never the validators inside one.
+        OrderField::Name => |_: &ValidatorRecord| None,
+        OrderField::StakeDelta7d => |_: &ValidatorRecord| None,
+        OrderField::StakeDelta30d => |_: &ValidatorRecord| None,
+        OrderField::Validators => |_: &ValidatorRecord| None,
```
ochaloup: 🤖 claude: Noting in code what @honzsik  raised in their review: these four extract `None` for every validator, so every comparison ties and the sort falls through to the unconditional `vote_account` tiebreak — `order_direction` is silently ignored, and `/validators?order_field=Name` returned a 400 before this change rather than an arbitrary order.

store/src/operators.rs
```diff
@@ -0,0 +1,41 @@
...
+        .comment(Some(b'#'))
+        .from_reader(OPERATORS_CSV.as_bytes())
+}
+
+fn registry() -> &'static HashMap<String, String> {
+    static REGISTRY: OnceLock<HashMap<String, String>> = OnceLock::new();
+    REGISTRY.get_or_init(|| {
+        let mut operators = HashMap::new();
+        let mut reader = reader();
+        for record in reader.records().flatten() {
```
ochaloup: 🤖 claude: Sorry if this was already settled — `.flatten()` discards any row whose field count is not 2 (an operator name containing a comma, say) with no log and no test, so that validator silently reports no operator and its group is short a member. Is that acceptable, or worth a `log::warn!` on the error arm instead?

api/src/handlers/list_validators.rs
```diff
@@ -129,68 +124,127 @@ pub async fn get_validators(
...
+/// Oldest epoch an `epochs`-wide window reaching back from `max_epoch` keeps.
+pub fn epoch_window_floor(max_epoch: u64, epochs: usize) -> u64 {
+    (max_epoch + 1).saturating_sub(epochs as u64)
+}
+
+pub fn trim_epoch_stats(validator: &mut ValidatorRecord, min_epoch: u64) {
+    validator
+        .epoch_stats
+        .retain(|stats| stats.epoch >= min_epoch);
 }
```
ochaloup: 🤖 claude: nit — both are one-expression wrappers with a single call site in this file, and `pub` here (as on `DEFAULT_EPOCHS`, `sort_validators`, `mod tests`, `epoch_stat` and `validator`) has no user anywhere outside it. If they are staged for a follow-up, a name that says what for would help the next reader; otherwise inlining them is the smaller change.

store/src/main.rs
```diff
@@ -55,9 +55,12 @@ pub mod close_epoch;
 pub mod cluster_info;
 pub mod commissions;
 pub mod dto;
+// Reached only through `utils`, which stamps operator ids and the incident window onto the records.
+pub mod groups;
```
ochaloup: 🤖 claude: nit — the binary only needs this module because `utils` reads `groups::INCIDENT_WINDOW_DAYS`; nothing here aggregates groups, and the comment credits `groups` with stamping operator ids, which `pub mod operators` below actually does. Moving that constant next to its only consumer in `utils`, under a name that states its unit, would drop this line and stop the binary compiling the whole aggregation module.

store/src/groups.rs
```diff
@@ -2,14 +2,18 @@ use crate::dto::{
...
 use chrono::{DateTime, Duration, Utc};
 use rust_decimal::prelude::*;
 use std::collections::{HashMap, HashSet};
 
 /// Windows the groups' stake deltas describe.
 const DELTA_SHORT_DAYS: i64 = 7;
 const DELTA_LONG_DAYS: i64 = 30;
+/// Window `incident_count_3m` describes. `load_incidents` reaches back 90 epochs, so this is a cut
+/// of what is already loaded rather than a second query.
+pub const INCIDENT_WINDOW_DAYS: i64 = 90;
```
ochaloup: 🤖 claude: This is 90 *days*, while `DEFAULT_INCIDENTS_WINDOW_EPOCHS` in `utils.rs` is 90 *epochs* (~180 days) — two constants, same number, different units, with only prose linking them. Names that carry the unit plus an assertion that the epoch window is the wider of the two would make narrowing that query safe; today it would silently shorten `incident_count_3m` with every test still green.

api/src/handlers/list_validators.rs
```diff
@@ -2,27 +2,32 @@ use std::cmp::Ordering;
...
-const DEFAULT_ORDER_FIELD: OrderField = OrderField::Stake;
-const DEFAULT_ORDER_DIRECTION: OrderDirection = OrderDirection::DESC;
 
 #[derive(Serialize, Debug, utoipa::ToSchema)]
 pub struct ResponseValidators {
     validators: Vec<ValidatorRecord>,
     validators_aggregated: Vec<ValidatorsAggregated>,
+    /// Operator rows, ordered by `order_field`, present only under `with_operator_groups`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    operators: Option<Vec<ValidatorGroupRecord>>,
```
ochaloup: 🤖 claude: Worth reconsidering this call — keeping only `.groups` drops `total_activated_stake` and `current_epoch`, which `/providers` and `/clients` both serve, and a consumer cannot recover the denominator by summing the rows because unmapped validators are excluded from them but counted in it.

api/src/cache.rs
```diff
@@ -155,8 +155,6 @@ impl Cache {
         self.validators.clone()
     }
 
-    // Aggregated once per refresh rather than per request: the rows are a few dozen, while the set
-    // they are derived from is every validator with every cached epoch.
```
ochaloup, line 159: 🤖 claude: This rationale looks worth keeping — it now covers three getters rather than one, and nothing else in the file states why the groups are aggregated once per refresh instead of per request. (The `{err}` → `{err:?}` change further down is unrelated to operator grouping too.)

ochaloup (commented):
I am adding few notes from the code perspective rather than the design&contract perspective as Honza did.

store/src/dto.rs
ilmari-h: yes reset this now

api/src/handlers/list_validators.rs
ilmari-h: improved this now, validators and operators are both now sortable by same columns

api/src/handlers/list_validators.rs
ilmari-h: inlining

store/src/operators.rs
ilmari-h: yes making this fail loudly on invalid format

api/src/cache.rs
ilmari-h: of course claude thinks it's own writing is worth keeping 😂
removed this because I thought it was fairly obvious as it's a member of `Cache`

api/src/handlers/list_validators.rs
ilmari-h: fixed

store/src/groups.rs

store/src/main.rs
ilmari-h: fixed

ilmari-h:
@honzsik thanks for those points. I now fixed those points.

Pagination applies to top-level rows (grouping in mind), sorting fixed, filtering applies also to group aggregate totals.

Asking for another check

ilmari-h:
@ochaloup  thanks for the review as well. Changed the implementation a bit as per honza's comments, re-requesting review.

api/src/handlers/list_validators.rs
```diff
@@ -234,12 +382,35 @@ fn get_field_extractor(order_field: OrderField) -> FieldExtractor {
...
+            SortKey::Text(
+                a.info_name
+                    .as_deref()
+                    .map(str::trim)
+                    .filter(|name| !name.is_empty())
+                    .unwrap_or(&a.vote_account)
+                    .to_lowercase(),
+            )
+        },
+        // A validator is one validator, so it ranks against an operator's count as such.
```
ochaloup: 🤖 claude: nit — this keys every validator identically, so on a plain `/validators?order_field=Validators` every row ties and `order_direction` has no effect. The value only means something under `with_operator_groups`, which is worth saying here.

```suggestion
        // A validator is one validator, so it ranks against an operator's count as such; without `with_operator_groups` every row ties and the direction does nothing.
```

api/src/handlers/list_validators.rs
```diff
@@ -127,105 +130,250 @@ pub async fn get_validators(
...
+            ),
+            ..aggregated
+        }
+    });
+    let (validators, total_count) = page_validators(
+        validators,
+        operators
+            .as_ref()
+            .map(|operators| operators.groups.as_slice()),
+        &config,
```
ochaloup, line 163: 🤖 claude: `page_validators` cuts the top-level rows, but `operators` comes back whole on every page, so a client rendering one header per row repeats all twelve operators on page 2 unless it cross-references the validators it received. Could this array be trimmed to the rows actually on the page, the way `page_tree` returns only its paged nodes?

store/src/groups.rs
```diff
@@ -260,22 +304,56 @@ fn aggregate_groups(
...
+            current_epoch,
+            eligible_stake,
         })
     }
 }
 
+/// Operator rows over just these validators, so a filtered list gets rows describing the validators
+/// it actually serves. `stake_share` is a share of them, not of the cluster.
+pub fn aggregate_operators(validators: &[&ValidatorRecord]) -> ValidatorGroups {
+    match Population::over(validators.to_vec(), validators.to_vec()) {
```
ochaloup, line 354: 🤖 claude: Passing the same slice as both rows and baselines means `group_stake_at` measures the reference epoch only over validators still in the list, so an operator whose member was unstaked to zero — and therefore dropped by `is_eligible_validator` before this is called — shows no loss in `stake_delta_7d` at all. The same validator's provider row does show it, because `Population::new` keeps ineligible records in `all`.

store/src/utils.rs
```diff
@@ -1372,11 +1381,26 @@ pub async fn load_validators(
...
+    let incidents_from = Utc::now() - chrono::Duration::days(INCIDENTS_COUNTED_DAYS);
     for (vote_account, record) in records.iter_mut() {
         record.incidents = incidents.get(vote_account).cloned().unwrap_or_default();
+        record.incident_count_3m = record
+            .incidents
+            .iter()
+            .filter(|incident| {
+                incident.start_at >= incidents_from
+                    && incident.downtime_seconds >= INCIDENTS_COUNTED_MIN_DOWNTIME_SECONDS
+            })
```
ochaloup, line 1325: 🤖 claude: This floor makes `incident_count_3m` stricter than `query_incident_free`, which defaults `min_incident_downtime_seconds` to 0 and applies no date window at all — so a validator with one 60-second restart last week reports 0 incidents and is still dropped by `?query_incident_free=true`. Defaulting that param to `INCIDENTS_COUNTED_MIN_DOWNTIME_SECONDS` and windowing the filter by `INCIDENTS_COUNTED_DAYS` would make the toggle mean exactly `incident_count_3m == 0`.

store/src/groups.rs
```diff
@@ -163,83 +211,79 @@ impl Accumulator {
...
+pub fn singleton_group(validator: &ValidatorRecord) -> ValidatorGroupRecord {
+    // Dropped the way `StakeWeighted::add` drops them.
+    let finite = |value: Option<f64>| value.filter(|value: &f64| value.is_finite());
+
+    ValidatorGroupRecord {
+        key: validator
+            .info_name
+            .clone()
+            .filter(|name| !name.trim().is_empty())
+            .unwrap_or_else(|| validator.vote_account.clone()),
```
ochaloup: 🤖 claude: `OrderField::Name` trims `info_name` before keying, this does not — so a validator named `" Acme"` gets the top-level key `" acme"`, and a leading space sorts before every letter, opening an ascending name sort above every operator while the list renders it as `Acme`.

```suggestion
        key: validator
            .info_name
            .as_deref()
            .map(str::trim)
            .filter(|name| !name.is_empty())
            .unwrap_or(&validator.vote_account)
            .to_string(),
```

api/src/handlers/list_validators.rs
```diff
@@ -127,105 +130,250 @@ pub async fn get_validators(
...
         .max()
         .unwrap_or(0);
-    let min_epoch = (max_epoch + 1).saturating_sub(config.epochs as u64);
+    // `epochs` counts the newest epoch itself.
+    let min_epoch = (newest_epoch + 1).saturating_sub(config.epochs as u64);
+
+    // Rows describe the validators this response serves, so they are aggregated per request.
+    let operators = config.with_operator_groups.then(|| {
+        let matching: Vec<&ValidatorRecord> = validators.iter().collect();
+        let aggregated = aggregate_operators(&matching);
```
ochaloup, line 149: 🤖 claude: Aggregating per request puts four walks of every cached epoch of every validator onto the Tokio worker — `eligible_stake`, `delta_epochs` and `group_stake_at` twice, then the accumulator loop — which is the class of work `warm_validators` deliberately moves into `spawn_blocking`. Worth measuring `with_operator_groups=true` under concurrency before assuming the per-request clone above already dominates it.

store/src/utils.rs
```diff
@@ -223,9 +223,14 @@ async fn get_apy_calculators(
...
-/// `incidents` field on `/validators`.
-const DEFAULT_INCIDENTS_WINDOW_EPOCHS: u64 = 90;
+/// Epochs `load_incidents` reaches back for the `incidents` array on `/validators`.
+const INCIDENTS_LOADED_EPOCHS: u64 = 90;
+
+/// Days `incident_count_3m` counts over, a cut of what `INCIDENTS_LOADED_EPOCHS` already loaded.
+pub const INCIDENTS_COUNTED_DAYS: i64 = 90;
+
+/// Downtime an interval needs to count towards `incident_count_3m`.
+pub const INCIDENTS_COUNTED_MIN_DOWNTIME_SECONDS: u64 = 180;
```
ochaloup, line 234: 🤖 claude: nit — both are `pub`, but their only readers outside this file are the `#[cfg(test)]` blocks in `store/src/groups.rs`, same crate; `INCIDENTS_LOADED_EPOCHS` three lines up is private for the same usage. `pub(crate)` unless the API crate is meant to read the floor — which would be one way to settle the `query_incident_free` mismatch below.

ochaloup (commented):
Sounds ok to me, just a few points from Claude review :-)
I think the most important is if the changes are aligned with FE.

api/src/handlers/list_validators.rs
ilmari-h: improved the comment

api/src/handlers/list_validators.rs
ilmari-h: this is nice, added

store/src/groups.rs
ilmari-h: 🤖 Confirmed for the case where a validator leaves the list entirely — `all == eligible` here, so `group_stake_at` reads the baseline over the same slice and the departed stake counts on neither side. One correction: it isn't unstaking that drops it. `is_eligible_validator` requires `activated_stake == 0 && credits == 0` over `last_epoch-1..=last_epoch`, so an unstaked-but-still-voting validator stays eligible and its loss does show; the invisible case is a full exit. Keeping it as is and documenting it, because these rows are aggregated per request over the query-filtered validators: cache-wide baselines would fold in stake from validators the request doesn't serve, and correct baselines mean splitting the eligibility filter from the query filters — while several of those filters (`query_sfdp`, `query_marinade_stake`, `query_superminority`) are stake-based and would exclude an exited validator anyway. Worth noting the two numbers aren't comparable by construction either way: operator deltas are per-request-filtered, client and provider deltas are cache-wide.

store/src/utils.rs
ilmari-h: Makes sense. Aligning both so they use the 3min minimum.

store/src/groups.rs
ilmari-h: done

store/src/utils.rs
ilmari-h: now read elsewhere also so leaving the `pub`s

api/src/handlers/list_validators.rs
ilmari-h: This handler is already entirely that class of work, with_operator_groups or not: cache.get_validators() clones the whole validator map (cache.rs:155), filter_validators walks all of it, sort_validators_ranked sorts it, and every response then clones and truncates epoch_stats per validator. Nothing in get_validators is on spawn_blocking today. So per-request aggregation isn't crossing a line the handler respects.

Measured impact and grouping seems to add about ~25% to request processing time, which is probably acceptable.

ilmari-h:
@ochaloup thank you for the comments! Fixed most and commented on rest. Ready for another one.

api/src/handlers/list_validators.rs
ochaloup: 🤖 claude: Confirmed — the trim is correct by construction, since a top-level row is wholly inside the rank window or wholly outside it. One leftover: the `operators` response field doc still describes the un-paged shape, so nothing public says the array is page-scoped.

store/src/groups.rs
ochaloup: 🤖 claude: Your correction is right and mine was wrong — `is_eligible_validator` keeps a validator with `activated_stake > 0 || credits > 0`, so an unstaking stays visible and only a full exit is not. Accepting the decision as documented; the `Population` struct doc above still states the opposite invariant unconditionally, which is worth a line now that the operator path is the exception.

store/src/utils.rs
ochaloup: 🤖 claude: Both halves landed, and the existing test was tightened with a boundary case rather than relaxed. One leftover: the `query_incident_free` doc still says it is evaluated over the last 90 epochs, where the filter now reads the last 90 days.

store/src/groups.rs
ochaloup: 🤖 claude: Applied, thanks. Nothing exercises a padded name on a standalone top-level row yet — the existing name test covers a blank name inside an operator block, which goes through `get_field_extractor` rather than `singleton_group`.

api/src/handlers/list_validators.rs
ochaloup: 🤖 claude: Fair — `get_validators` was never on `spawn_blocking`, so per-request aggregation is not crossing a line the handler respects. Thanks for measuring; noting the ~25% here as the baseline if it ever needs revisiting.

store/src/utils.rs
ochaloup: 🤖 claude: Agreed — `filter_validators` reads both constants now, so the `pub` is load-bearing rather than accidental. That was the alternative the comment named, so nothing further here.

dualisimo (commented):
Filtered findings from Claude which sounds reasonable:

- `store/src/groups.rs:255` — Unknown commission/uptime flips a validator from one end of the list to the other when `with_operator_groups` is toggled: `singleton_group` leaves them `None` → `SortKey::Missing` (always last), while `get_field_extractor` substitutes worst-case sentinels (commission → 100, uptime → 0).
- [list_validators.rs:456](api/src/handlers/list_validators.rs:456) — `query` never matches `validator.operator`, so `?with_operator_groups=true&query=Coinbase` returns nothing despite a Coinbase operator block existing. `record.operator` is already stamped before filtering.
- `store/src/stake_deltas.rs:29` — `delta_epochs` anchors on the newest `epoch_end_at` rather than `now`; since the in-progress epoch has no end, the reference epoch is one epoch older than master picked, so `stake_delta_7d` spans ~9–11 days under a 7-day label (30d shifts likewise).
- `store/src/operators.rs:33` — A vote account listed under two operators is silently last-wins, in a parser otherwise built to fail loudly on bad input. An `assert!(insert(..).is_none())` would close it.

api/src/handlers/list_validators.rs
```diff
@@ -1,29 +1,48 @@
...
+    /// operator rows' `stake_share`. Summing the rows does not recover it: a validator belonging to no
+    /// operator counts here and has no row. Present only under `with_operator_groups`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    total_activated_stake: Option<Decimal>,
+    /// Epoch the operator rows describe.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    current_epoch: Option<u64>,
+    /// Number of rows matching the query and filters, before `offset`/`limit`: validators, or
+    /// top-level rows under `with_operator_groups`.
     total_count: usize,
```
kronael, line 46: 🤖 `total_count` changes unit under `with_operator_groups` — validators without it, top-level rows with it — and so do `offset`/`limit`. A client that flips the flag and keeps its pager gets wrong page math with no type error and no failing test.

The doc comment is the minimum. A distinct `total_rows` field (or `total_count` staying validators, with rows reported separately) would make it unmissable.

api/src/handlers/list_validators.rs
```diff
@@ -50,39 +69,20 @@ pub struct QueryParams {
...
+    /// Minimum downtime in seconds for a `DOWN` interval to count as an incident for `query_incident_free`. Shorter intervals are restart noise. Defaults to the floor `incident_count_3m` uses, and like that count only intervals from the last 90 days are read. Ignored unless `query_incident_free` is set, and never filters the returned `incidents` array.
     min_incident_downtime_seconds: Option<u64>,
     query_verified: Option<bool>,
     query_protected: Option<bool>,
     query_flagged: Option<bool>,
     /// When true, `query` also matches datacenter location fields (country, city) in addition to
     /// validator name, vote account and identity.
     search_properties: Option<bool>,
+    /// `true` groups the validators into operator blocks, returns the `operators` aggregates beside them, and pages over those top-level rows rather than validators.
+    with_operator_groups: Option<bool>,
```
kronael, line 81: 🤖 The part worth pushing back on is not the paging unit but the fact that a flag changes the response **shape**.

`/validators` now answers two different questions — a flat validator list, and an operator-aggregated drilldown — behind one path, one handler, and one response type with three conditionally-absent fields. Every client has to branch on a request parameter to know what it got; the OpenAPI schema describes a union it cannot express; and `get_validators` carries a second sort/rank/page path (`top_level_ranks`, `sort_validators_ranked`, the `retain` in `page_validators`) that only runs half the time.

For analytical drilldowns like this, GraphQL is the better fit and likely the cheaper one: the client names the aggregation it wants, so the shape lives in the query instead of in a boolean, and the server can skip work nobody asked for rather than running `aggregate_operators` over the full match on every flagged request. Splitting into two REST endpoints (`/validators`, `/operators`) would fix the shape problem too — GraphQL additionally fixes the field-level over-fetching this handler already has (15 epochs of `epoch_stats` on every row, whatever the caller needs).

kronael (dismissed):

api/src/handlers/list_validators.rs
ilmari-h: yes this was a bit of an unorthodox way to do this but to serve the frontend's needs without complicating the codebase, I felt adding a couple more fields to the response was a reasonable choice. The response shape does change but only additions: it stays bw compatible and the caller knows what to expect.

api/src/handlers/list_validators.rs
ilmari-h: Yes the pagination logic changes and so does what this field represents. This is to serve the pagination-needs of the frontend.

kronael (approved):

PR #262, opened 2026-08-21
[GEN-8375] fix null validator version bug

ilmari-h:
Top-level version was seeded from the newest epoch row alone, so a gossip gap in the most recent epochs made a validator read as versionless despite older epochs holding a version. Added a fill-down over the epoch-DESC rows so the record takes the newest observed version, leaving epoch_stats[].version null where nothing was observed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validator records now consistently display the latest available client and version metadata, including RPC, pubsub, and gossip details.
  * Metadata is only applied to the correct validator identity, improving accuracy when records overlap or contain gaps.
  * Per-epoch validator statistics retain metadata specific to each epoch, preventing older records from overwriting newer details.
  * Validator metadata remains correctly populated when information is missing or appears across different epochs.

### DIFF
--- a/store/src/utils.rs
+++ b/store/src/utils.rs
@@ -1201,6 +1201,11 @@ pub async fn load_validators(
                     rugged_commission_occurrences: 0,
                 });
 
+            // Gossip can miss a node for a whole epoch, leaving that epoch's row versionless; rows arrive epoch-DESC, so this lands the newest epoch that observed a version.
+            if record.version.is_none() {
+                record.version = row.get("version");
+            }
+
             let seeding_epoch = *seeding_epochs.entry(vote_account.clone()).or_insert(epoch);
             // Gating all three on max_observed keeps them from one row; stopping one epoch below the record's own is what makes that row the newest closed epoch rather than whichever older row still happens to have the column. commission_advertised deliberately stays on the open epoch that seeded the record.
             if record.commission_max_observed.is_none() && epoch + 1 >= seeding_epoch {
--- a/store/src/utils.rs
+++ b/store/src/utils.rs
@@ -1207,17 +1207,18 @@ pub async fn load_validators(
                 });
 
             // Rows are newest-first, so project all node metadata from the first usable row.
-            let has_node_metadata = [
-                rpc_public.is_some(),
-                version.is_some(),
-                stored_client_id.is_some(),
-                client_id_raw.is_some(),
-                feature_set.is_some(),
-                shred_version.is_some(),
-            ]
-            .into_iter()
-            .any(|present| present);
-            if has_node_metadata && !projected_node_metadata.contains(&vote_account) {
+            let has_node_metadata = rpc_public.is_some()
+                || pubsub_public.is_some()
+                || gossip_port.is_some()
+                || version.is_some()
+                || stored_client_id.is_some()
+                || client_id_raw.is_some()
+                || feature_set.is_some()
+                || shred_version.is_some();
+            if has_node_metadata
+                && row.get::<_, &str>("identity") == record.identity
+                && !projected_node_metadata.contains(&vote_account)
+            {
                 record.version = version.clone();
                 record.client_id = client_id;
                 record.client_name = client_name(client_id);
@@ -1295,9 +1296,9 @@ pub async fn load_validators(
                 client_id_raw,
                 feature_set,
                 shred_version,
-                gossip_port: row.get::<_, Option<i32>>("gossip_port").map(|n| n as u16),
+                gossip_port,
                 rpc_public,
-                pubsub_public: row.get("pubsub_public"),
+                pubsub_public,
                 activated_stake: row.get::<_, Decimal>("activated_stake"),
                 marinade_stake: row.get::<_, Decimal>("marinade_stake"),
                 foundation_stake: row.get::<_, Decimal>("foundation_stake"),
--- a/store/src/utils.rs
+++ b/store/src/utils.rs
@@ -1071,6 +1071,7 @@ pub async fn load_validators(
         log::info!("Aggregating validator records...");
         let mut records: HashMap<_, _> = Default::default();
         let mut seeding_epochs: HashMap<String, u64> = Default::default();
+        let mut projected_node_metadata: HashSet<String> = Default::default();
         for row in rows {
             let vote_account: String = row.get("vote_account");
             let epoch: u64 = row.get::<_, Decimal>("epoch").try_into().unwrap();
@@ -1122,11 +1123,13 @@ pub async fn load_validators(
                 .as_ref()
                 .and_then(|c| c.dc_concentration_by_country.get(&dc_country).cloned());
 
+            let version: Option<String> = row.get("version");
+            let stored_client_id = row.get::<_, Option<i32>>("client_id").map(|n| n as u16);
             let client_id_raw: Option<String> = row.get("client_id_raw");
-            let client_id = effective_client_id(
-                row.get::<_, Option<i32>>("client_id").map(|n| n as u16),
-                client_id_raw.as_deref(),
-            );
+            let client_id = effective_client_id(stored_client_id, client_id_raw.as_deref());
+            let feature_set = row.get::<_, Option<i64>>("feature_set").map(|n| n as u32);
+            let shred_version = row.get::<_, Option<i32>>("shred_version").map(|n| n as u16);
+            let rpc_public: Option<bool> = row.get("rpc_public");
 
             let record = records
                 .entry(vote_account.clone())
@@ -1158,17 +1161,17 @@ pub async fn load_validators(
                     commission_advertised: row.get::<_, Option<i32>>("commission_advertised"),
                     commission_effective: row.get::<_, Option<i32>>("commission_effective"),
                     commission_aggregated: None,
-                    version: row.get("version"),
+                    version: version.clone(),
                     client_id,
                     client_name: client_name(client_id),
                     client_label: client_label(client_id),
                     client_vendor: client_vendor(client_id),
                     client_lineage: client_lineage(client_id),
                     client_id_raw: client_id_raw.clone(),
-                    feature_set: row.get::<_, Option<i64>>("feature_set").map(|n| n as u32),
-                    shred_version: row.get::<_, Option<i32>>("shred_version").map(|n| n as u16),
+                    feature_set,
+                    shred_version,
                     gossip_port: row.get::<_, Option<i32>>("gossip_port").map(|n| n as u16),
-                    rpc_public: row.get("rpc_public"),
+                    rpc_public,
                     pubsub_public: row.get("pubsub_public"),
                     activated_stake: row.get::<_, Decimal>("activated_stake"),
                     marinade_stake: row.get::<_, Decimal>("marinade_stake"),
@@ -1201,9 +1204,28 @@ pub async fn load_validators(
                     rugged_commission_occurrences: 0,
                 });
 
-            // Gossip can miss a node for a whole epoch, leaving that epoch's row versionless; rows arrive epoch-DESC, so this lands the newest epoch that observed a version.
-            if record.version.is_none() {
-                record.version = row.get("version");
+            // Rows are newest-first, so project all node metadata from the first usable row.
+            let has_node_metadata = [
+                rpc_public.is_some(),
+                version.is_some(),
+                stored_client_id.is_some(),
+                client_id_raw.is_some(),
+                feature_set.is_some(),
+                shred_version.is_some(),
+            ]
+            .into_iter()
+            .any(|present| present);
+            if has_node_metadata && !projected_node_metadata.contains(&vote_account) {
+                record.version = version.clone();
+                record.client_id = client_id;
+                record.client_name = client_name(client_id);
+                record.client_label = client_label(client_id);
+                record.client_vendor = client_vendor(client_id);
+                record.client_lineage = client_lineage(client_id);
+                record.client_id_raw = client_id_raw.clone();
+                record.feature_set = feature_set;
+                record.shred_version = shred_version;
+                projected_node_metadata.insert(vote_account.clone());
             }
 
             let seeding_epoch = *seeding_epochs.entry(vote_account.clone()).or_insert(epoch);
@@ -1253,7 +1275,7 @@ pub async fn load_validators(
                 commission_effective: row
                     .get::<_, Option<i32>>("commission_effective")
                     .map(|n| n.try_into().unwrap()),
-                version: row.get("version"),
+                version,
                 mev_commission_bps: row.get::<_, Option<i32>>("mev_commission_bps"),
                 priority_commission_bps: row.get::<_, Option<i32>>("priority_commission_bps"),
                 dc_asn: row.get::<_, Option<i32>>("dc_asn"),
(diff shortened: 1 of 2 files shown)


store/src/utils.rs
```diff
@@ -1201,6 +1201,11 @@ pub async fn load_validators(
                     rugged_commission_occurrences: 0,
                 });
 
+            // Gossip can miss a node for a whole epoch, leaving that epoch's row versionless; rows arrive epoch-DESC, so this lands the newest epoch that observed a version.
+            if record.version.is_none() {
+                record.version = row.get("version");
+            }
```
ochaloup: 🤖 claude: `display_epochs` is `DEFAULT_CACHE_EPOCHS` (80) on the API path, so this can take a version from 60 epochs back while `/validators?epochs=10` returns ten `epoch_stats` entries that are all versionless — SFDP then reads a months-old version as the node's current one, with nothing in the response to locate it. The commission fill-down right below bounds itself with `epoch + 1 >= seeding_epoch` for exactly this reason; a bound here would keep a version that no longer describes a live node from being served.

store/src/utils.rs
```diff
@@ -1201,6 +1201,11 @@ pub async fn load_validators(
                     rugged_commission_occurrences: 0,
                 });
 
+            // Gossip can miss a node for a whole epoch, leaving that epoch's row versionless; rows arrive epoch-DESC, so this lands the newest epoch that observed a version.
+            if record.version.is_none() {
+                record.version = row.get("version");
```
ochaloup: 🤖 claude: `client_id_raw`, `client_name`, `feature_set` and `shred_version` come from the same gossip observation and go null in the same epochs — the writer `COALESCE`s `version` and `client_id_raw` identically — so Jaguar will now report `version: 4.1.0-rc.1` alongside `client_name: Unknown` and `feature_set: null`, a combination no epoch observed. `load_validators_aggregated_flat` deliberately resolves version and the client columns off one row; the whole gossip group should move together here too.

store/src/utils.rs
ochaloup: Isn't this what `row.get("version")` is returned in the same way as the `if` added at the bottom?

ochaloup (commented):
This simple change is a bit strange to me - I'm not sure why `row.get` wasn't enough in the original record creation. Then Claude came up with a strangely "large" number of points. Can you please check?

store/src/utils.rs
ilmari-h: This only inserts it if it didn't already exist.

Bug is the entry exists in `records` but it's version there is `None`. So this `or_insert_with` is never exercised after the record for that vote account is inserted once. If that happens to have `None` version, the version stays as `None`. 
Rows are ordered by epoch descending, so this selects the newest non-null version.

store/src/utils.rs
ilmari-h: fixed, the fix now applies to whole gossip group's fields

store/src/utils.rs
ilmari-h: fixed in recent commit

store/src/utils.rs
```diff
@@ -1201,6 +1204,30 @@ pub async fn load_validators(
...
+                record.version = version.clone();
+                record.client_id = client_id;
+                record.client_name = client_name(client_id);
+                record.client_label = client_label(client_id);
+                record.client_vendor = client_vendor(client_id);
+                record.client_lineage = client_lineage(client_id);
+                record.client_id_raw = client_id_raw.clone();
+                record.feature_set = feature_set;
+                record.shred_version = shred_version;
+                projected_node_metadata.insert(vote_account.clone());
```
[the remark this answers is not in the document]

store/src/utils.rs
```diff
@@ -1201,6 +1206,33 @@ pub async fn load_validators(
...
+                record.client_label = client_label(client_id);
+                record.client_vendor = client_vendor(client_id);
+                record.client_lineage = client_lineage(client_id);
+                record.client_id_raw = client_id_raw.clone();
+                record.feature_set = feature_set;
+                record.shred_version = shred_version;
+                record.gossip_port = gossip_port;
+                record.rpc_public = rpc_public;
+                record.pubsub_public = pubsub_public;
+                projected_node_metadata.insert(vote_account.clone());
```
ochaloup, line 1234: 🤖 claude: Note this has already been reported — and looking at the replies, the "fixed in recent commit" seems to have been meant for the sibling thread about the gossip group moving together, which genuinely is fixed. This one is still open: nothing bounds how far back the first usable row may sit, so with `DEFAULT_CACHE_EPOCHS = 80` on the API path a `/validators?epochs=15` response can carry metadata from 80 epochs ago with nothing in it to say so, while the commission fill-down three lines below still gates itself with `epoch + 1 >= seeding_epoch`.

Maybe this was an intention to leave it as it is and not adjust it. So let me know if that's the case.

store/src/utils.rs
```diff
@@ -1201,6 +1206,33 @@ pub async fn load_validators(
...
+                rpc_public.is_some(),
+                version.is_some(),
+                stored_client_id.is_some(),
+                client_id_raw.is_some(),
+                feature_set.is_some(),
+                shred_version.is_some(),
+            ]
+            .into_iter()
+            .any(|present| present);
+            if has_node_metadata && !projected_node_metadata.contains(&vote_account) {
```
ochaloup: 🤖 claude: `identity` is a per-epoch column (`validators` is keyed on `(vote_account, epoch)`) but `record.identity` is seeded from the newest row only, and every field projected here is collected per identity via `node_info.get(&identity)` in `collect/src/validators.rs`. So a vote account that rotates its node identity, where the new identity is not in gossip yet, gets the old node's version, client, feature set and ports reported under the new identity — the same "combination no epoch observed" this block set out to remove, moved from across-epochs to across-identities.

```suggestion
            if has_node_metadata
                && row.get::<_, &str>("identity") == record.identity
                && !projected_node_metadata.contains(&vote_account)
            {
```

store/src/utils.rs
```diff
@@ -1201,6 +1206,33 @@ pub async fn load_validators(
...
+            let has_node_metadata = [
+                rpc_public.is_some(),
+                version.is_some(),
+                stored_client_id.is_some(),
+                client_id_raw.is_some(),
+                feature_set.is_some(),
+                shred_version.is_some(),
+            ]
+            .into_iter()
+            .any(|present| present);
```
ochaloup: 🤖 claude: nit — this is a boolean disjunction written as an array plus an identity closure, which also evaluates all six terms instead of short-circuiting.

```suggestion
            let has_node_metadata = rpc_public.is_some()
                || version.is_some()
                || stored_client_id.is_some()
                || client_id_raw.is_some()
                || feature_set.is_some()
                || shred_version.is_some();
```

store/src/utils.rs
```diff
@@ -1201,6 +1206,33 @@ pub async fn load_validators(
...
+                record.client_name = client_name(client_id);
+                record.client_label = client_label(client_id);
+                record.client_vendor = client_vendor(client_id);
+                record.client_lineage = client_lineage(client_id);
+                record.client_id_raw = client_id_raw.clone();
+                record.feature_set = feature_set;
+                record.shred_version = shred_version;
+                record.gossip_port = gossip_port;
+                record.rpc_public = rpc_public;
+                record.pubsub_public = pubsub_public;
```
ochaloup, line 1233: 🤖 claude: nit — `pubsub_public` and `gossip_port` are projected here but are not among the six presence checks that decide whether to project, so the gate is narrower than what it assigns. It happens to be safe only because `collect/src/validators.rs` derives both from `node.map(..)`/`node.and_then(..)`, making `rpc_public.is_some()` a proxy for "node was in gossip" — worth listing all eight so that cross-crate invariant stops being load-bearing.

store/src/utils.rs
```diff
@@ -1261,10 +1293,10 @@ pub async fn load_validators(
...
                 client_lineage: client_lineage(client_id),
                 client_id_raw,
-                feature_set: row.get::<_, Option<i64>>("feature_set").map(|n| n as u32),
-                shred_version: row.get::<_, Option<i32>>("shred_version").map(|n| n as u16),
+                feature_set,
+                shred_version,
                 gossip_port: row.get::<_, Option<i32>>("gossip_port").map(|n| n as u16),
-                rpc_public: row.get("rpc_public"),
+                rpc_public,
                 pubsub_public: row.get("pubsub_public"),
```
ochaloup: 🤖 claude: nit — these two still re-read from the row although the hoisted locals are in scope and `Copy`, so the block reads as if they were deliberately sourced differently from `feature_set`, `shred_version` and `rpc_public` next to them.

```suggestion
                gossip_port,
                rpc_public,
                pubsub_public,
```

store/tests/validator_version_projection_sql.rs
```diff
@@ -0,0 +1,155 @@
...
+    assert_eq!(seen.client_lineage, Some("agave".to_string()));
+    assert_eq!(seen.client_id_raw, Some("Jito".to_string()));
+    assert_eq!(seen.feature_set, Some(333));
+    assert_eq!(seen.shred_version, Some(444));
+    assert_eq!(seen.rpc_public, Some(true));
+    assert_eq!(seen.pubsub_public, Some(false));
+    assert_eq!(seen.gossip_port, Some(8001));
+
+    let partial = validators.get("votePartial").unwrap();
+    assert_eq!(partial.version, None);
```
ochaloup, line 127: 🤖 claude: nit — this is the most consequential assertion in the file and the only surprising one without a message: it locks in that a node *seen in gossip* with no version reports `version: null` even when an older epoch has one, because `rpc_public.is_some()` latches the projection on that versionless row. Defensible and consistent with the `retained` case, but it means only a full gossip gap fills down, so it is worth saying out loud here.

```suggestion
    assert_eq!(
        partial.version,
        None,
        "a node seen in gossip reporting no version has no version; only a full gossip gap fills down"
    );
```

store/tests/validator_version_projection_sql.rs
```diff
@@ -0,0 +1,155 @@
...
+    assert_eq!(partial.client_id, Some(1));
+    assert_eq!(partial.client_name, "Jito Labs");
+    assert_eq!(partial.feature_set, Some(333));
+    assert_eq!(partial.shred_version, Some(444));
+    assert_eq!(partial.rpc_public, Some(false));
```
ochaloup, line 132: 🤖 claude: nit — this assertion cannot fail: both `votePartial` rows carry `rpc_public = false`, so it passes whichever row the projection picks (the same reason CodeRabbit flagged `pubsub_public` above, which is `false` in both rows of both fixtures). Two more gaps while you are here: `gap.epoch_stats[1]`, the second of the two blank epochs, is never asserted, and `voteNeverSeen` checks only `version`, not that the rest of the group also stays `None`.

ochaloup (commented):
@ilmari-h I'm sorry for a bit of ping-pong here. The claude returned me a list of findings that I think could be reasoable to verify. Posting them down here.

store/src/utils.rs
ilmari-h: the 80 epoch load window is existing behavior so not changing it here.

store/tests/validator_version_projection_sql.rs
ilmari-h: leaving test nit out of scope

store/tests/validator_version_projection_sql.rs
ilmari-h: ignoring test nit

store/src/utils.rs

store/src/utils.rs
ilmari-h: fixed

store/src/utils.rs

store/src/utils.rs

store/tests/validator_version_projection_sql.rs
```diff
@@ -0,0 +1,155 @@
...
+                ('identityGossipGap', 'voteGossipGap', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
+                ('identitySeen', 'voteSeen', $2, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.0.3', 3, 'Agave', 111, 222, false, false, 8000),
+                ('identitySeen', 'voteSeen', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.1.0', 1, 'Jito', 333, 444, true, false, 8001),
+                ('identityPartial', 'votePartial', $2, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.0.3', 3, 'Agave', 111, 222, false, false, 8000),
+                ('identityPartial', 'votePartial', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, 1, 'Jito', 333, 444, false, false, 8001),
+                ('identityRetained', 'voteRetained', $2, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.0.3', 3, 'Agave', 111, 222, false, false, 8000),
+                ('identityRetained', 'voteRetained', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.1.0', 1, 'Jito', NULL, NULL, NULL, NULL, NULL),
+                ('identityNeverSeen', 'voteNeverSeen', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL)",
```
ochaloup, line 48: 🤖 claude: every fixture here uses a single identity per vote account, so the new `row.identity == record.identity` guard is unconditionally true — deleting that clause outright leaves all 12 store SQL tests green, which means the branch deciding whether the fix applies at all is unpinned. A rotated-identity fixture closes that, together with `assert_eq!(rotated.version, None, "a node that changed identity must not inherit the previous node's software")` and `assert_eq!(rotated.gossip_port, None)` next to the `voteNeverSeen` assertions.

```suggestion
                ('identityNeverSeen', 'voteNeverSeen', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
                ('identityOld', 'voteRotated', $1, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), '4.0.3', 3, 'Agave', 111, 222, false, false, 8000),
                ('identityNew', 'voteRotated', $2, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
                ('identityNew', 'voteRotated', $3, 100, 0, 0, false, 0, 0, 0, 0, 0, NOW(), NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL)",
```

store/src/utils.rs
```diff
@@ -1201,6 +1206,34 @@ pub async fn load_validators(
...
+            let has_node_metadata = rpc_public.is_some()
+                || pubsub_public.is_some()
+                || gossip_port.is_some()
+                || version.is_some()
+                || stored_client_id.is_some()
+                || client_id_raw.is_some()
+                || feature_set.is_some()
+                || shred_version.is_some();
+            if has_node_metadata
+                && row.get::<_, &str>("identity") == record.identity
```
ochaloup, line 1219: 🤖 claude: nit — `identity` is now read from the row twice, here and where `ValidatorRecord.identity` is seeded, while every other column this block touches is hoisted into a local once, so the block is half-converted again for the same reason the `gossip_port` / `pubsub_public` reads were. Hoisting `let identity: &str = row.get("identity");` next to the other locals and using `identity.to_owned()` in the `or_insert_with` closure keeps both borrows of `row` immutable and compiles unchanged.

ochaloup (approved):
On the added code claude comes with some nits. They are kind of relevant but feel free to ignore them and merge the PR.

PR #263, opened 2026-08-26
[GEN-8802] [trivial] increase epoch stats window 80->90 to match incidents window

ilmari-h:
## Summary by CodeRabbit

* **Bug Fixes**
  * Validator event queries now use the last 90 epochs by default, providing a broader history window.
  * Increased the default cached epoch range to 90 epochs for improved consistency with incident data.
  * Block production statistics now include the first epoch in the requested range.

### DIFF
--- a/api/src/handlers/events.rs
+++ b/api/src/handlers/events.rs
@@ -16,7 +16,7 @@ pub struct ResponseEvents {
 #[derive(Deserialize, Serialize, Debug, utoipa::IntoParams)]
 #[into_params(parameter_in = Query)]
 pub struct QueryParams {
-    /// Lower-bound epoch (inclusive). Mutually exclusive with `query_from_date`. Defaults to the last 80 epochs.
+    /// Lower-bound epoch (inclusive). Mutually exclusive with `query_from_date`. Defaults to the last 90 epochs.
     query_from_epoch: Option<u64>,
     /// Lower-bound date (RFC3339), resolved to the first epoch ending on/after it. Mutually exclusive with `query_from_epoch`.
     query_from_date: Option<DateTime<Utc>>,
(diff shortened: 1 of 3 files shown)


api/src/handlers/events.rs
```diff
@@ -16,7 +16,7 @@ pub struct ResponseEvents {
 #[derive(Deserialize, Serialize, Debug, utoipa::IntoParams)]
 #[into_params(parameter_in = Query)]
 pub struct QueryParams {
-    /// Lower-bound epoch (inclusive). Mutually exclusive with `query_from_date`. Defaults to the last 80 epochs.
+    /// Lower-bound epoch (inclusive). Mutually exclusive with `query_from_date`. Defaults to the last 90 epochs.
```
ochaloup, line 19: 🤖 claude: The default is 91 epochs, not 90 — `get_events_with_context` (`store/src/validators_events.rs:136`) subtracts `DEFAULT_CACHE_EPOCHS` and then queries `epoch >= $2`, while every other consumer of the constant uses the strict `epoch > last_epoch - $1`, so `/validators/{vote_account}/events` returns one epoch that `/validators` `epoch_stats` does not cover. Fix with `.saturating_add(1).saturating_sub(DEFAULT_CACHE_EPOCHS)`, the idiom already used in `api/src/handlers/list_validators.rs:139`.

store/src/utils.rs
```diff
@@ -227,6 +227,9 @@ async fn get_apy_calculators(
 /// `incidents` field on `/validators`.
 const DEFAULT_INCIDENTS_WINDOW_EPOCHS: u64 = 90;
 
+// Keep default cached epochs at least the size of the incidents window.
```
ochaloup, line 230: 🤖 claude: nit — the comment transcribes the inequality on the next line; what is not derivable from the code is the consequence of the invariant breaking.

```suggestion
// Otherwise /validators reports incidents for epochs its epoch_stats no longer covers.
```

ochaloup (changes requested):
I believe the comment for the `assert!` should explain "why" rather than "what".

Otherwise the +1 on paginating seems not such an important point :-)

api/src/handlers/events.rs
ilmari-h: fixed

ochaloup (approved):

PR #265, opened 2026-08-27
[GEN-8631] add expected take rates to the take-rates historical data

ilmari-h:
Introducing the expected take rate into the take rate history endpoint in addition to the realized one.

Renamed field `take_rate` to `realized_take_rate` on purpose. No need for bw compat, api not yet used by anyone.

## Summary by CodeRabbit

* **New Features**
  * Added expected take-rate calculations based on each epoch’s reward composition.
  * Take-rate results now distinguish realized and expected rates.
  * Added support for inflation, MEV, priority-fee, and block-reward shares.
* **Performance**
  * Improved take-rate data retrieval with caching and database indexing.
* **Bug Fixes**
  * Improved handling of missing reward data, unknown accounts, commission limits, and block-only rewards.
* **Documentation**
  * Clarified expected take-rate descriptions and API response fields.

### DIFF
--- a/api.md
+++ b/api.md
@@ -145,6 +145,9 @@ Query parameters, mutually exclusive; without either one the whole stored histor
 
 `epoch_start_at` and `epoch_end_at` are `null` for epochs whose boundaries are not recorded.
 Such an epoch is still returned by both query modes, including the open one.
+
+`realized_take_rate` is what the validator kept that epoch; `expected_take_rate` is what its
+commissions that epoch implied it would, `null` where no commission was recorded.
 ```bash
 curl -sfLS localhost:8000/validators/XkCriyrNwS3G4rzAXtG5B1nnvb5Ka1JtCku93VqeKAr/take-rates | jq
 ```
@@ -155,7 +158,8 @@ curl -sfLS localhost:8000/validators/XkCriyrNwS3G4rzAXtG5B1nnvb5Ka1JtCku93VqeKAr
       "epoch": 378,
       "epoch_start_at": "2022-11-27T19:45:05.669098Z",
       "epoch_end_at": "2022-11-29T19:45:05.669098Z",
-      "take_rate": 0.0812,
+      "realized_take_rate": 0.0812,
+      "expected_take_rate": 0.1052,
       "created_at": "2022-11-30T15:58:04.038843Z"
     }
   ]
--- a/api.md
+++ b/api.md
@@ -147,7 +147,8 @@ Query parameters, mutually exclusive; without either one the whole stored histor
 Such an epoch is still returned by both query modes, including the open one.
 
 `realized_take_rate` is what the validator kept that epoch; `expected_take_rate` is what its
-commissions that epoch implied it would, `null` where no commission was recorded.
+commissions that epoch implied it would. It is `null` where no commission was recorded, and for
+the in-progress epoch, which has paid no inflation rewards to weight them by.
 ```bash
 curl -sfLS localhost:8000/validators/XkCriyrNwS3G4rzAXtG5B1nnvb5Ka1JtCku93VqeKAr/take-rates | jq
 ```
--- a/store/src/dto.rs
+++ b/store/src/dto.rs
@@ -483,7 +483,9 @@ pub struct TakeRateRecord {
     pub epoch: u64,
     pub epoch_start_at: Option<DateTime<Utc>>,
     pub epoch_end_at: Option<DateTime<Utc>>,
-    pub take_rate: f64,
+    pub realized_take_rate: f64,
+    /// This epoch's commissions weighted by this epoch's reward mix, against `realized_take_rate`'s measured figure. Inflation side is the worse of the epoch's `commission_max_observed` and `commission_advertised`, so any mid-epoch move reads high. Not the same window as `/validators`'s `expected_take_rate`, which pairs the current rate with a trailing 30-day mix. Null when the epoch has no commission recorded.
+    pub expected_take_rate: Option<f64>,
     pub created_at: DateTime<Utc>,
 }
 
--- a/store/src/dto.rs
+++ b/store/src/dto.rs
@@ -484,7 +484,7 @@ pub struct TakeRateRecord {
     pub epoch_start_at: Option<DateTime<Utc>>,
     pub epoch_end_at: Option<DateTime<Utc>>,
     pub realized_take_rate: f64,
-    /// This epoch's commissions weighted by this epoch's reward mix, against `realized_take_rate`'s measured figure. Inflation side is the worse of the epoch's `commission_max_observed` and `commission_advertised`, so any mid-epoch move reads high. Not the same window as `/validators`'s `expected_take_rate`, which pairs the current rate with a trailing 30-day mix. Null when the epoch has no commission recorded.
+    /// Take rate implied by the epoch's commissions, weighted by the epoch's reward mix.
     pub expected_take_rate: Option<f64>,
     pub created_at: DateTime<Utc>,
 }
--- a/store/src/take_rates.rs
+++ b/store/src/take_rates.rs
@@ -201,15 +201,14 @@ pub async fn load_epoch_reward_mix(
         if total.is_zero() {
             continue;
         }
+        let total = total.to_f64().unwrap_or_default();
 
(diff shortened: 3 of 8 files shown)


store/src/take_rates.rs
```diff
@@ -192,13 +257,25 @@ pub async fn get_take_rate_series(
...
+                expected_take_rate(
+                    *shares,
+                    worst_known_commission(
+                        row.get::<_, Option<i32>>("commission_max_observed"),
+                        row.get::<_, Option<i32>>("commission_advertised"),
+                    ),
+                    row.get::<_, Option<i32>>("mev_commission_bps"),
+                    row.get::<_, Option<i32>>("priority_commission_bps"),
+                )
+            }),
```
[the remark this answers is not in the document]
ilmari-h: The precondition is unreachable. shares.inflation == 0.0 requires SUM(inflation_rewards) == 0 cluster-wide for an epoch. Inflation is the one reward component no validator can opt out of

dualisimo (commented):
Filtered findings from Claude which sounds reasonable:

- [store/src/take_rates.rs:247](store/src/take_rates.rs:247) — Epochs with no Jito distribution-account snapshot read too high: `priority_commission_bps = NULL` makes `expected_take_rate` count the whole block share as kept, and `mev` gaps drop MEV from the denominator. Those rows can't be backfilled (the collector scrapes live program accounts; Jito closes them after the claim window) while `validators_rewards` backfills arbitrarily far, so the series silently mixes epochs of very different data quality. `/validators` dodges this with its 10-epoch `DEFAULT_JITO_COMMISSION_EPOCHS` window — the historical path drops that protection.
- [store/src/take_rates.rs:182](store/src/take_rates.rs:182) — `load_epoch_reward_mix` includes the in-progress epoch, whose row can hold only accruing `block_rewards`; the mix becomes ≈`{0, 0, 1}`, the inflation commission gets a zero weight, and a non-Jito validator's newest point reads 100%.
- [api.md:158](api.md:158) — Sample response still documents `take_rate`, not `realized_take_rate`, and omits `expected_take_rate`.
- [store/src/dto.rs:487](store/src/dto.rs:487) — Doc says null means "no commission recorded", but null also covers an unwarmed reward-mix cache (which the new test asserts) — indistinguishable to callers.

store/src/dto.rs
```diff
@@ -483,7 +483,9 @@ pub struct TakeRateRecord {
     pub epoch: u64,
     pub epoch_start_at: Option<DateTime<Utc>>,
     pub epoch_end_at: Option<DateTime<Utc>>,
-    pub take_rate: f64,
+    pub realized_take_rate: f64,
+    /// This epoch's commissions weighted by this epoch's reward mix, against `realized_take_rate`'s measured figure. Inflation side is the worse of the epoch's `commission_max_observed` and `commission_advertised`, so any mid-epoch move reads high. Not the same window as `/validators`'s `expected_take_rate`, which pairs the current rate with a trailing 30-day mix. Null when the epoch has no commission recorded.
```
kronael: is this describing what it is - as in definition - or some analytical reasoning?  Imho should be just the former clearly stated.

store/src/take_rates.rs
```diff
@@ -168,20 +169,84 @@ pub async fn store_take_rates(
...
+        let mev: Decimal = row.get("mev");
+        let block: Decimal = row.get("block");
+
+        let total = inflation + mev + block;
+        if total.is_zero() {
+            continue;
+        }
+
+        // Divided in Decimal: cluster-wide lamport sums reach f64's exact-integer ceiling.
+        let share = |component: Decimal| (component / total).to_f64().unwrap_or_default();
```
kronael: Is this needed? https://docs.rs/rust_decimal/latest/src/rust_decimal/decimal.rs.html#2552

kronael (dismissed):

store/src/take_rates.rs
ilmari-h: thought it made it a bit more readable but kinda noisy, fixing

store/src/dto.rs
ilmari-h: trimmed it, it was AI written

ilmari-h:
thanks @kronael fixed those points, asking for another review

ilmari-h:
> Filtered findings from Claude which sounds reasonable:
> 
> * [store/src/take_rates.rs:247](store/src/take_rates.rs:247) — Epochs with no Jito distribution-account snapshot read too high: `priority_commission_bps = NULL` makes `expected_take_rate` count the whole block share as kept, and `mev` gaps drop MEV from the denominator. Those rows can't be backfilled (the collector scrapes live program accounts; Jito closes them after the claim window) while `validators_rewards` backfills arbitrarily far, so the series silently mixes epochs of very different data quality. `/validators` dodges this with its 10-epoch `DEFAULT_JITO_COMMISSION_EPOCHS` window — the historical path drops that protection.
> * [store/src/take_rates.rs:182](store/src/take_rates.rs:182) — `load_epoch_reward_mix` includes the in-progress epoch, whose row can hold only accruing `block_rewards`; the mix becomes ≈`{0, 0, 1}`, the inflation commission gets a zero weight, and a non-Jito validator's newest point reads 100%.
> * api.md:158 — Sample response still documents `take_rate`, not `realized_take_rate`, and omits `expected_take_rate`.
> * [store/src/dto.rs:487](store/src/dto.rs:487) — Doc says null means "no commission recorded", but null also covers an unwarmed reward-mix cache (which the new test asserts) — indistinguishable to callers.

Thanks, first point is ok, this depends on what epochs we collect. validator rewards epochs never reach lower than mev collected epochs so this wont be a problem.

2nd and 3rd good findings thanks, fixed now in recent commit

api.md
```diff
@@ -145,6 +145,9 @@ Query parameters, mutually exclusive; without either one the whole stored histor
 
 `epoch_start_at` and `epoch_end_at` are `null` for epochs whose boundaries are not recorded.
 Such an epoch is still returned by both query modes, including the open one.
+
+`realized_take_rate` is what the validator kept that epoch; `expected_take_rate` is what its
+commissions that epoch implied it would, `null` where no commission was recorded.
```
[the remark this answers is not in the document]
ilmari-h: done

store/src/utils.rs
```diff
@@ -785,13 +785,19 @@ pub async fn load_take_rates() -> anyhow::Result<TakeRates> {
...
+/// Inflation is the one component no validator can opt out of.
 pub fn expected_take_rate(
     shares: RewardMixShares,
     inflation_commission_pct: Option<i32>,
     mev_commission_bps: Option<i32>,
     priority_commission_bps: Option<i32>,
 ) -> Option<f64> {
+    // An in-progress epoch has paid no inflation or MEV yet, leaving a mix of pure block rewards.
+    if shares.inflation <= 0.0 {
+        return None;
```
kronael, line 798: is this diffferent from =0?  Just asking so that if yes, we could get rid of the Option here.

store/src/utils.rs
```diff
@@ -785,13 +785,19 @@ pub async fn load_take_rates() -> anyhow::Result<TakeRates> {
...
-/// None when the inflation commission is unknown, which is the one component no validator can opt out of.
+/// None when the inflation commission is unknown, or when the mix carries no inflation to weight it by.
+/// Inflation is the one component no validator can opt out of.
 pub fn expected_take_rate(
     shares: RewardMixShares,
     inflation_commission_pct: Option<i32>,
     mev_commission_bps: Option<i32>,
     priority_commission_bps: Option<i32>,
 ) -> Option<f64> {
+    // An in-progress epoch has paid no inflation or MEV yet, leaving a mix of pure block rewards.
```
kronael, line 796: how can an in-progress epoch have inflation rewards?

kronael (approved):
some nits

store/src/utils.rs
ilmari-h: its excluding the in-progress/current epoch there (which has no inflation rewards) or a previous epoch where the data about inflation rewards hasn't landed yet

store/src/utils.rs
ilmari-h: Yep, we don't know the take rate yet so it's None. Inflation lands at next epoch boundary, so then this would go from 0->X, so saying its 0 is not correct.

PR #266, opened 2026-08-28
[GEN-8735]  load operators csv from config file pointed to by env var

ilmari-h:
Load operator group config from a file at runtime. Makes it configurable via ops infra.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved operator CSV validation and error reporting for missing headers, malformed records, and blank fields.
  * Errors now identify the affected source file for easier troubleshooting.
* **Documentation**
  * Clarified references to the example operator CSV file.
* **Chores**
  * Added local test configuration using the example operator data.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+# `OPERATORS_CSV` is required at runtime and has no in-code default. Cargo sets it here so tests
+# and local runs resolve; the container sets the real one, and a binary run outside cargo fails
+# loudly when it is unset.
+[env]
+OPERATORS_CSV = { value = "store/operators.example.csv", relative = true }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,0 @@
-# `OPERATORS_CSV` is required at runtime and has no in-code default. Cargo sets it here so tests
-# and local runs resolve; the container sets the real one, and a binary run outside cargo fails
-# loudly when it is unset.
-[env]
-OPERATORS_CSV = { value = "store/operators.example.csv", relative = true }
--- a/store/operators.default.csv
+++ b/store/operators.default.csv
@@ -1,3 +1,4 @@
+# Stale snapshot; up-to-date config lives in ops-infra. Override with OPERATORS_CSV.
 vote_account,operator
 
 # Binance
--- a/store/src/operators.rs
+++ b/store/src/operators.rs
@@ -2,32 +2,47 @@ use crate::dto::ValidatorRecord;
 use std::collections::HashMap;
 use std::sync::OnceLock;
 
-const OPERATORS_CSV: &str = include_str!("../operators.csv");
-
 fn registry() -> &'static HashMap<String, String> {
     static REGISTRY: OnceLock<HashMap<String, String>> = OnceLock::new();
-    REGISTRY.get_or_init(|| parse(OPERATORS_CSV))
+    REGISTRY.get_or_init(|| {
+        let path = std::env::var_os("OPERATORS_CSV")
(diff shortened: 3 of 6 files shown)


store/src/operators.rs
```diff
@@ -2,32 +2,47 @@ use crate::dto::ValidatorRecord;
...
         let (Some(vote_account), Some(operator)) = (record.get(0), record.get(1)) else {
-            panic!("operators.csv: {record:?} is not a vote_account,operator pair");
+            panic!("{source}: {record:?} is not a vote_account,operator pair");
         };
 
         let (vote_account, operator) = (vote_account.trim(), operator.trim());
         assert!(
             !vote_account.is_empty() && !operator.is_empty(),
-            "operators.csv: {record:?} leaves the vote account or the operator blank"
+            "{source}: {record:?} leaves the vote account or the operator blank"
```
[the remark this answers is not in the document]
ilmari-h: out of scope here

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+# `OPERATORS_CSV` is required at runtime and has no in-code default. Cargo sets it here so tests
+# and local runs resolve; the container sets the real one, and a binary run outside cargo fails
+# loudly when it is unset.
+[env]
+OPERATORS_CSV = { value = "store/operators.example.csv", relative = true }
```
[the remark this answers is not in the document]
ilmari-h: this is fine to act as the text fixture

kronael: what is relative=true?

kronael: also a default might just be a default in code or set in the test, why in .cargo?

store/src/operators.rs
```diff
@@ -2,32 +2,47 @@ use crate::dto::ValidatorRecord;
...
+            .unwrap_or_else(|err| panic!("OPERATORS_CSV {path:?}: {err}"));
+        parse(&path.to_string_lossy(), &csv)
+    })
 }
 
 /// `#` starts a comment; the file groups its rows under one per operator.
 ///
 /// Panics if format wrong.
-fn parse(csv: &str) -> HashMap<String, String> {
+fn parse(source: &str, csv: &str) -> HashMap<String, String> {
```
kronael: I'd like to propose an alternative (this is good too, just while thinking about it).

a) parse returns Result - anyhow
b) caller adds context or enriches the error by a generic "parse {file} failed: {err}"

This way the inner function does not care about path, yet all errors are rich enough to debug easily.

store/operators.default.csv
kronael, line 1: I would probably imagine doing operators changes in a special PR, but if these are just to be treated as "tests", then its ok to have it bundled here.

kronael (dismissed):
Some suggestions, but its ok as it is.  Thank you.

.cargo/config.toml
ilmari-h: i think relative to the dir where this is ran from as opposed to absolute directory path, this is just for tests

ilmari-h: that is a good point, let me see if I can improve this...

store/operators.default.csv
ilmari-h: yeah operators changes are in ops-infra and the prod ones are driven from there. This just changes the default/example file to align with that.

.cargo/config.toml
ilmari-h: yeah now I just baked it in as a static string by default, with comments that this is a stale snapshot, real one driven from ops-infra via config file. This is cleaner I think

store/src/operators.rs
ilmari-h: agreed, done

kronael (approved):

PR #267, opened 2026-08-28
[GEN-8735] make returned incidents agree with query param

ilmari-h:
`query_incident_free` would return records that had `incidents` array entries. Making these now agree on incident epoch window and what counts as an incident.

Removed field `incident_count_3m` - this was AI noise that somehow made it to my prev PR. The validator incidents are now in `incidents` array only.

The /clients and /providers endpoint don't carry the incidents count sort option anymore because not needed by FE and would result in very large responses potentially for large provider and client groups.

## Summary by CodeRabbit

* **New Features**
  * Validator listings support configurable incident time windows and downtime thresholds.
  * Validator groups expose detailed incident records, including timing and downtime.
  * Incident data is included in the API documentation.

* **Bug Fixes**
  * Sorting clients or providers by incidents now returns a clear validation error.
  * Incident filtering, ordering, and aggregation consistently reflect retained records and configured windows.
  * Invalid incident windows are rejected with an explanatory error.
  * Restart-related noise and incidents below the downtime threshold are excluded from results.

### DIFF
--- a/api/src/handlers/list_clients.rs
+++ b/api/src/handlers/list_clients.rs
@@ -1,10 +1,8 @@
 use crate::context::WrappedContext;
 use crate::metrics;
 use crate::utils::order::{
-    incidents_order_rejection, OrderDirection, OrderField, DEFAULT_ORDER_DIRECTION,
-    DEFAULT_ORDER_FIELD,
+    OrderDirection, OrderField, DEFAULT_ORDER_DIRECTION, DEFAULT_ORDER_FIELD,
 };
-use crate::utils::response::response_error;
 use crate::utils::validator_groups::{page_tree, GetGroupsConfig, DEFAULT_LIMIT};
 use chrono::{DateTime, Utc};
 use rust_decimal::prelude::*;
@@ -51,10 +49,6 @@ pub async fn handler(
     context: WrappedContext,
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_CLIENTS.inc();
-
-    if let Some(message) = incidents_order_rejection(query_params.order_field, "clients") {
-        return Ok(response_error(StatusCode::BAD_REQUEST, message));
-    }
 
     let config = GetGroupsConfig {
         order_field: query_params.order_field.unwrap_or(DEFAULT_ORDER_FIELD),
--- a/api/src/handlers/list_clients.rs
+++ b/api/src/handlers/list_clients.rs
@@ -1,7 +1,8 @@
 use crate::context::WrappedContext;
 use crate::metrics;
 use crate::utils::order::{
-    OrderDirection, OrderField, DEFAULT_ORDER_DIRECTION, DEFAULT_ORDER_FIELD,
+    incidents_order_rejection, OrderDirection, OrderField, DEFAULT_ORDER_DIRECTION,
+    DEFAULT_ORDER_FIELD,
 };
 use crate::utils::response::response_error;
 use crate::utils::validator_groups::{page_tree, GetGroupsConfig, DEFAULT_LIMIT};
@@ -51,11 +52,8 @@ pub async fn handler(
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_CLIENTS.inc();
 
-    if query_params.order_field == Some(OrderField::Incidents) {
-        return Ok(response_error(
-            StatusCode::BAD_REQUEST,
-            "order_field=incidents is not available for clients".into(),
-        ));
+    if let Some(message) = incidents_order_rejection(query_params.order_field, "clients") {
+        return Ok(response_error(StatusCode::BAD_REQUEST, message));
     }
 
     let config = GetGroupsConfig {
--- a/api/src/handlers/list_clients.rs
+++ b/api/src/handlers/list_clients.rs
@@ -3,6 +3,7 @@ use crate::metrics;
 use crate::utils::order::{
     OrderDirection, OrderField, DEFAULT_ORDER_DIRECTION, DEFAULT_ORDER_FIELD,
 };
+use crate::utils::response::response_error;
 use crate::utils::validator_groups::{page_tree, GetGroupsConfig, DEFAULT_LIMIT};
 use chrono::{DateTime, Utc};
 use rust_decimal::prelude::*;
@@ -49,6 +50,13 @@ pub async fn handler(
     context: WrappedContext,
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_CLIENTS.inc();
+
+    if query_params.order_field == Some(OrderField::Incidents) {
+        return Ok(response_error(
+            StatusCode::BAD_REQUEST,
+            "order_field=incidents is not available for clients".into(),
+        ));
+    }
 
     let config = GetGroupsConfig {
         order_field: query_params.order_field.unwrap_or(DEFAULT_ORDER_FIELD),
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -534,11 +534,11 @@ pub async fn handler(
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_VALIDATORS.inc();
     if let Some(window) = query_params.incident_window_epochs {
-        if window == 0 || window > DEFAULT_INCIDENTS_WINDOW_EPOCHS {
+        if window == 0 || window > DEFAULT_CACHE_EPOCHS {
             return Ok(response_error(
                 StatusCode::BAD_REQUEST,
                 format!(
-                    "incident_window_epochs must be between 1 and {DEFAULT_INCIDENTS_WINDOW_EPOCHS}"
+                    "incident_window_epochs must be between 1 and {DEFAULT_CACHE_EPOCHS}"
                 ),
             ));
         }
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -5,23 +5,21 @@ use crate::metrics;
 use crate::utils::order::{
     compare_keys, OrderDirection, OrderField, SortKey, DEFAULT_ORDER_DIRECTION, DEFAULT_ORDER_FIELD,
 };
-use crate::utils::response::response_error_500;
+use crate::utils::response::{response_error, response_error_500};
 use crate::utils::validator_groups::{compare_group_rows, group_column, sort_groups};
-use chrono::{DateTime, Duration, Utc};
+use chrono::{DateTime, Utc};
 use log::error;
 use rust_decimal::prelude::*;
 use serde::{Deserialize, Serialize};
 use store::{
     dto::{ValidatorGroupRecord, ValidatorGroups, ValidatorRecord, ValidatorsAggregated},
     groups::{aggregate_operators, singleton_group},
-    utils::{
-        to_fixed_for_sort, worst_known_commission, INCIDENTS_COUNTED_DAYS,
-        INCIDENTS_COUNTED_MIN_DOWNTIME_SECONDS,
-    },
+    utils::{to_fixed_for_sort, worst_known_commission, INCIDENTS_WINDOW_EPOCHS},
 };
 use warp::{http::StatusCode, reply::json, Reply};
 
 const DEFAULT_EPOCHS: usize = 15;
+const DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS: u64 = 180;
 const DEFAULT_LIMIT: usize = 100;
 
 #[derive(Serialize, Debug, utoipa::ToSchema)]
@@ -67,10 +65,12 @@ pub struct QueryParams {
     query_marinade_stake: Option<bool>,
     query_with_names: Option<bool>,
     query_sfdp: Option<bool>,
-    /// Evaluated over the last 90 epochs of incidents, regardless of `epochs` and `query_from_date`.
+    /// Evaluated over `incident_window_epochs` of incidents, regardless of `epochs` and `query_from_date`.
     query_incident_free: Option<bool>,
-    /// Minimum downtime in seconds for a `DOWN` interval to count as an incident for `query_incident_free`. Shorter intervals are restart noise. Defaults to the floor `incident_count_3m` uses, and like that count only intervals from the last 90 days are read. Ignored unless `query_incident_free` is set, and never filters the returned `incidents` array.
+    /// Minimum downtime in seconds for a `DOWN` interval to read as an incident. Shorter intervals are restart noise: they are dropped from the returned `incidents` array, from what `order_field=incidents` ranks on and from what `query_incident_free` reads. Applies whether or not `query_incident_free` is set.
     min_incident_downtime_seconds: Option<u64>,
+    /// Epochs back `query_incident_free` reads incidents over, counting the newest epoch itself. Defaults to the whole window the `incidents` array carries; 400 above it, since older incidents are not loaded. Ignored unless `query_incident_free` is set, and never trims the returned `incidents` array.
+    incident_window_epochs: Option<u64>,
     query_verified: Option<bool>,
     query_protected: Option<bool>,
     query_flagged: Option<bool>,
@@ -99,6 +99,7 @@ pub struct GetValidatorsConfig {
     pub query_sfdp: Option<bool>,
     pub query_incident_free: Option<bool>,
     pub min_incident_downtime_seconds: Option<u64>,
+    pub incident_window_epochs: Option<u64>,
     pub query_verified: Option<bool>,
     pub query_protected: Option<bool>,
     pub query_flagged: Option<bool>,
@@ -403,7 +404,7 @@ fn get_field_extractor(order_field: OrderField) -> FieldExtractor {
             |a: &ValidatorRecord| a.unique_delegators.map(Decimal::from).into()
         }
         OrderField::Incidents => {
-            |a: &ValidatorRecord| SortKey::Number(Decimal::from(a.incident_count_3m))
+            |a: &ValidatorRecord| SortKey::Number(Decimal::from(a.incidents.len()))
         }
         OrderField::StakeDelta7d => |a: &ValidatorRecord| a.stake_delta_7d.into(),
         OrderField::StakeDelta30d => |a: &ValidatorRecord| a.stake_delta_30d.into(),
@@ -431,6 +432,17 @@ pub fn filter_validators(
     // Shared with the client and provider aggregates, so both describe the same population.
     let last_epoch = store::utils::last_reported_epoch(validators.values()).unwrap_or(0);
     validators.retain(|_, validator| store::utils::is_eligible_validator(validator, last_epoch));
+
+    // Everything downstream reads whatever survives this floor: the array served, the ordering,
+    // `query_incident_free`, and the operator rows aggregated off these records.
+    let min_incident_downtime = config
+        .min_incident_downtime_seconds
+        .unwrap_or(DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS);
+    for validator in validators.values_mut() {
+        validator
+            .incidents
+            .retain(|incident| incident.downtime_seconds >= min_incident_downtime);
+    }
 
     if config.query_sfdp.is_some() {
         validators.retain(|_, validator| validator.foundation_stake.gt(&Decimal::ZERO))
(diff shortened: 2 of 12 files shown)


api/src/handlers/list_validators.rs
```diff
@@ -478,15 +490,14 @@ pub fn filter_validators(
...
+                .incident_window_epochs
+                .unwrap_or(INCIDENTS_WINDOW_EPOCHS),
+        );
         validators.retain(|_, v| {
-            let has_incident = v.incidents.iter().any(|i| {
-                i.start_at >= incidents_from && i.downtime_seconds >= min_incident_downtime
-            });
+            let has_incident = v.incidents.iter().any(|i| i.epoch >= from_epoch);
             has_incident != query_incident_free
         });
```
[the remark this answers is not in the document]
ilmari-h: good one, changing it so that `query_incident_free` agrees with `incidents` on what the epoch window is, controlled by `incident_window_epochs`

api/src/handlers/list_clients.rs
```diff
@@ -50,6 +51,13 @@ pub async fn handler(
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_CLIENTS.inc();
 
+    if query_params.order_field == Some(OrderField::Incidents) {
```
kronael, line 54: nit: would always write this as if let ... but don't really see why I would. 🤔

kronael, line 54: Why is it even valid in the schema then, if its not available?

api/src/utils/order.rs
```diff
@@ -23,6 +23,7 @@ pub enum OrderField {
     ExpectedTakeRate,
     Validators,
     DelegationRelationships,
+    /// Only `/validators` rows carry incidents; `/clients` and `/providers` answer 400 on it.
```
kronael, line 26: nit: this is a type system abuse making it rather not do its job as it should

store/src/groups.rs
```diff
@@ -26,6 +26,12 @@ impl GroupKind {
     fn drops_unclassified(self) -> bool {
         matches!(self, GroupKind::Operator)
     }
+
+    /// Only operator rows carry their members' incidents. A client or provider row spans most of
```
kronael: nit: is this comment insightful?

kronael (dismissed):

api/src/handlers/list_clients.rs
ilmari-h: this type is shared with /validators where it IS available. And this used to be available before this.

didn't want to duplicate code. Plan is to enable this later also for /clients and /providers but for now disabling with this runtime gate. Just didn't wanna include it in this PR as it could cause some performance issue and not sure how I would implement it there.

api/src/utils/order.rs
ilmari-h: agreed, see above comment

store/src/groups.rs
ilmari-h: I guess claude thought it was 😂 
but a bit redundant

ilmari-h:
thanks for the review @kronael - fixed conflicts with master and removed one AI slop type comment, asking for another review

kronael (dismissed):

store/src/groups.rs
```diff
@@ -260,7 +288,11 @@ pub fn singleton_group(validator: &ValidatorRecord) -> ValidatorGroupRecord {
         uptime_pct: finite(validator.avg_uptime_pct),
         expected_take_rate: finite(validator.expected_take_rate),
         delegation_relationship_count: validator.unique_delegators,
-        incident_count_3m: validator.incident_count_3m,
+        incidents: validator
+            .incidents
+            .iter()
+            .map(|incident| GroupIncidentRecord::new(&validator.vote_account, incident))
+            .collect(),
```
ochaloup, line 295: 🤖 claude: `singleton_group`'s own doc says the row "is only ever sorted, never served", yet this builds a `GroupIncidentRecord` — each with a heap `String` for the vote account — for every incident of every operator-less validator, on every request, only so `group_column` can read `.len()`. `top_level_ranks` walks the whole matched set rather than the page, so with `min_incident_downtime_seconds=0` (no floor, restart noise included) that is on the order of 100k values built and dropped per call, where the old `incident_count_3m` was a `u64` copy.

api/src/handlers/list_clients.rs
```diff
@@ -50,6 +51,13 @@ pub async fn handler(
...
+        return Ok(response_error(
+            StatusCode::BAD_REQUEST,
+            "order_field=incidents is not available for clients".into(),
+        ));
+    }
```
ochaloup, line 59: 🤖 claude: None of the three new 400 branches has a test — this one, the identical rejection in `list_providers.rs`, and `incident_window_epochs` being `0` or above `90` in `list_validators.rs`. If any of them regresses, the endpoint silently sorts every row on an always-empty vector instead of rejecting the request, which is exactly the confusion the check exists to prevent.

api/src/handlers/list_validators.rs
```diff
@@ -521,6 +533,16 @@ pub async fn handler(
     context: WrappedContext,
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_VALIDATORS.inc();
+    if let Some(window) = query_params.incident_window_epochs {
+        if window == 0 || window > DEFAULT_INCIDENTS_WINDOW_EPOCHS {
+            return Ok(response_error(
+                StatusCode::BAD_REQUEST,
+                format!(
+                    "incident_window_epochs must be between 1 and {DEFAULT_INCIDENTS_WINDOW_EPOCHS}"
```
ochaloup, line 541: 🤖 claude: `DEFAULT_INCIDENTS_WINDOW_EPOCHS` is doing double duty here as both the default window and the accepted maximum, but the real ceiling is what the cache actually holds — `DEFAULT_CACHE_EPOCHS`, which is what the const assert above expresses. Tune the default down to 30 for product reasons and `incident_window_epochs=60` starts answering 400 despite the cache still holding 90 epochs, with an error message that reads self-consistent so nothing signals the regression.

```suggestion
        if window == 0 || window > DEFAULT_CACHE_EPOCHS {
            return Ok(response_error(
                StatusCode::BAD_REQUEST,
                format!(
                    "incident_window_epochs must be between 1 and {DEFAULT_CACHE_EPOCHS}"
```

store/src/dto.rs
```diff
@@ -592,7 +589,31 @@ pub struct ValidatorGroupRecord {
     pub uptime_pct: Option<f64>,
     pub expected_take_rate: Option<f64>,
     pub delegation_relationship_count: Option<u64>,
-    pub incident_count_3m: u64,
+    /// Every member's incidents, oldest first. Two members down at once are two records here.
```
ochaloup, line 592: 🤖 claude: `GroupKind::carries_incidents` is true only for `Operator`, so `/clients` and `/providers` always serialize `incidents: []` — but this doc is the schema utoipa publishes for all three endpoints, and it promises every member's incidents. A consumer generated from it reads the empty array on `/providers` as "no hosting provider had downtime in the window"; the caveat currently lives only on `OrderField::Incidents`.

```suggestion
    /// Every member's incidents, oldest first; only operator rows populate this, `/clients` and `/providers` always answer empty. Two members down at once are two records here.
```

store/src/groups.rs
```diff
@@ -130,10 +134,18 @@ struct Accumulator {
     uptime_pct: StakeWeighted,
     expected_take_rate: StakeWeighted,
     delegation_relationship_count: Option<u64>,
-    incident_count_3m: u64,
+    carries_incidents: bool,
+    incidents: Vec<GroupIncidentRecord>,
```
ochaloup, line 138: 🤖 claude: nit — `Accumulator` still derives `Default`, which sets `carries_incidents: false`, the wrong value for the one kind that carries them. Making it work required changing the single construction site from `.or_default()` to `.or_insert_with(|| Accumulator::new(kind))`; storing `kind: GroupKind` and dropping the derive instead would make a future second accumulation path impossible to get silently wrong — operator rows coming back with `incidents: []` would not even compile.

store/src/dto.rs
```diff
@@ -390,7 +387,7 @@ pub struct UptimeRecord {
 }
 
 /// A single downtime incident (one DOWN interval from the uptimes table).
-#[derive(Deserialize, Serialize, Debug, Clone, utoipa::ToSchema)]
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, utoipa::ToSchema)]
```
ochaloup, line 390: 🤖 claude: nit — this `PartialEq` is unused: removing it and running `cargo build --all-targets` across the workspace still compiles clean, since `ValidatorRecord` does not derive `PartialEq` and the tests that build incidents only need `Clone`. The sibling derive on `GroupIncidentRecord` is genuinely required, because `ValidatorGroupRecord` derives `PartialEq`.

```suggestion
#[derive(Deserialize, Serialize, Debug, Clone, utoipa::ToSchema)]
```

store/src/utils.rs
```diff
@@ -224,44 +224,30 @@ async fn get_apy_calculators(
...
             SELECT
                 vote_account,
                 uptimes.epoch,
                 start_at,
                 end_at,
                 EXTRACT('epoch' FROM (end_at - start_at))::BIGINT AS downtime_seconds
             FROM uptimes
-            CROSS JOIN cluster
-            WHERE status = 'DOWN' AND uptimes.epoch > cluster.last_epoch - $1::NUMERIC
+            WHERE status = 'DOWN' AND uptimes.epoch >= $1::NUMERIC
```
ochaloup, line 248: 🤖 claude: This query has no test coverage at all — `store/tests/` holds 14 SQL tests on the shared harness and none of them touches `uptimes`. The uncovered case is precisely the one the new comment in `load_validators` cites as the reason for the rewrite: a `cluster_info` head sitting an epoch either side of the `validators` head would shift the served incident window, and the handler's default window silently disagrees with the loader, with nothing to catch it.

ochaloup (commented):
I believe the functionality is here fine. I am just placing down few points from the Claude feedback while they do not seem critical.

store/src/groups.rs
ilmari-h: comment was out of date

api/src/handlers/list_validators.rs
ilmari-h: done

store/src/dto.rs
ilmari-h: changed comment

store/src/dto.rs
ilmari-h: 🤖 confirmed and removed — workspace builds clean without it.

store/src/groups.rs
ilmari-h: 🤖 done — `Accumulator` stores `kind: GroupKind` and the `Default` derive is gone, so it cannot be constructed without one. `carries_incidents` is now read off the kind at the point it is used.

store/src/utils.rs
ilmari-h: 🤖 added `store/tests/incidents_sql.rs` on the shared harness — five tests: the window starts on `from_epoch` itself (the off-by-one that would make the loader and the handler disagree), an UP interval is not an incident, `downtime_seconds` is the interval length, each DOWN row is its own incident oldest first, and rows are keyed by the validator that was down. Ran green against a real Postgres, not just the skip path.

The `cluster_info`-vs-`validators` head case itself is not covered — that lives in `load_validators`, which needs a full fixture set across a dozen tables to reach. What is covered is the query it feeds, so a change to the bound or the ordering now fails a test.

api/src/handlers/list_clients.rs
ilmari-h: 🤖 partly. The `/clients` and `/providers` rejection is now one `incidents_order_rejection` in `api/src/utils/order.rs` that both handlers call, with tests covering both group names and the columns left alone.

Skipped the `incident_window_epochs` one. Making it testable meant extracting a helper for what is a two-condition range check, and that trade did not look worth it — the handler itself is not unit-testable either way, since `WrappedContext` needs a live `tokio_postgres::Client`.

ilmari-h:
@ochaloup fixed the suggestions, asking for another review

api/src/handlers/list_validators.rs
```diff
@@ -67,10 +69,12 @@ pub struct QueryParams {
...
+    /// `true` keeps the validators whose `incidents` array comes back empty, `false` the rest. It reads that array, so `min_incident_downtime_seconds` and `incident_window_epochs` shape it too, where `epochs` and `query_from_date` do not.
     query_incident_free: Option<bool>,
-    /// Minimum downtime in seconds for a `DOWN` interval to count as an incident for `query_incident_free`. Shorter intervals are restart noise. Defaults to the floor `incident_count_3m` uses, and like that count only intervals from the last 90 days are read. Ignored unless `query_incident_free` is set, and never filters the returned `incidents` array.
+    /// Minimum downtime in seconds for a `DOWN` interval to read as an incident. Shorter intervals are restart noise, and reach neither the `incidents` array nor `order_field=incidents` nor `query_incident_free`.
     min_incident_downtime_seconds: Option<u64>,
+    /// Epochs back the `incidents` array reaches, counting the newest reported epoch itself. Defaults to 90, the whole window the cache holds, and answers 400 above it. Unrelated to `epochs`, which sizes `epoch_stats`.
```
ochaloup: 🤖 claude: The check now bounds on `DEFAULT_CACHE_EPOCHS` while the default stays `DEFAULT_INCIDENTS_WINDOW_EPOCHS`, but this doc still attaches the 400 to the default — tune the default to 30 and the published schema says "defaults to 90, 400 above it" while the API still accepts 90.

```suggestion
    /// Epochs back the `incidents` array reaches, counting the newest reported epoch itself. Defaults to 90; above 90 — the whole window the cache holds — answers 400. Unrelated to `epochs`, which sizes `epoch_stats`.
```

store/src/groups.rs
```diff
@@ -130,10 +136,29 @@ struct Accumulator {
...
+            net_apy: Default::default(),
+            take_rate: Default::default(),
+            credits: Default::default(),
+            marinade_score: Default::default(),
+            apy: Default::default(),
+            commission: Default::default(),
+            uptime_pct: Default::default(),
+            expected_take_rate: Default::default(),
+            delegation_relationship_count: None,
+            incidents: Vec::new(),
```
ochaloup: 🤖 claude: nit — dropping the derive was the right call, but the literal now spells "empty" three ways: `Default::default()`, explicit `0` / `Decimal::ZERO` / `None`, and `Vec::new()`. One idiom across the twelve empty fields would leave `kind` as the only line carrying a value, which is the property the comment above the struct is claiming.

store/src/utils.rs
ochaloup: 🤖 claude: The window is open at the top, and `uptimes` is written every minute (`collector-performance`, `schedule: '* * * * *'`) while `validators` — where `last_epoch` comes from — is written hourly (`collector-validators`, `schedule: 0 * * * *`), so for up to an hour after each epoch rollover `MAX(uptimes.epoch)` sits one above the anchor and `incident_window_epochs=W` serves W+1 epochs. Either pass `last_epoch` in as a second parameter and add `AND uptimes.epoch <= $2::NUMERIC`, or state on the parameter doc that the newer epoch is served deliberately — both it and `the_window_starts_on_from_epoch_itself`'s comment currently promise a closed window.

store/src/groups.rs
```diff
@@ -222,14 +254,22 @@ impl Accumulator {
...
+                incidents
+            },
         }
     }
 }
 
 /// The row a validator belonging to no group stands for on its own. Ordered against the aggregated
-/// rows, so it has to read the same fields `Accumulator::add` does. `stake_share` is left at zero;
-/// the row is only ever sorted, never served.
+/// rows, so it has to read the same fields `Accumulator::add` does.
```
ochaloup, line 262: 🤖 claude: Both halves of the deleted sentence still hold — `stake_share: 0.0` is hardcoded eight lines down, and `top_level_ranks` is still the only non-test caller, keeping just the extracted `SortKey` and the name — so the zero now has no explanation left in the code.

```suggestion
/// rows, so it has to read the same fields `Accumulator::add` does. `stake_share` is left at zero;
/// the row is only ever sorted, never served.
```

store/src/groups.rs
```diff
@@ -260,7 +300,11 @@ pub fn singleton_group(validator: &ValidatorRecord) -> ValidatorGroupRecord {
         uptime_pct: finite(validator.avg_uptime_pct),
         expected_take_rate: finite(validator.expected_take_rate),
         delegation_relationship_count: validator.unique_delegators,
-        incident_count_3m: validator.incident_count_3m,
+        incidents: validator
+            .incidents
+            .iter()
+            .map(|incident| GroupIncidentRecord::new(&validator.vote_account, incident))
+            .collect(),
```
ochaloup: 🤖 claude: Before this change the standalone row carried `incident_count_3m: validator.incident_count_3m` — a `u64` copy — and `group_column` read it directly; now `top_level_ranks` walks the whole matched set rather than the page, and every operator-less validator gets one `GroupIncidentRecord` per incident, each heap-allocating a 44-char vote-account `String`, only for `.len()` at `validator_groups.rs:75` before the record is dropped (it keeps just the `SortKey` and the name). With `min_incident_downtime_seconds=0` the restart-noise floor is gone, so a single caller-controlled request builds and throws away on the order of 100k `String`s.

What it should be: leave `incidents` empty here and rank the standalone case on `validator.incidents.len()` at the `singleton_group` call site in `top_level_ranks`, or give `group_column` a length-only path for `OrderField::Incidents` — either restores the O(1), allocation-free sort key the `u64` gave for free. Note the two halves must land together: emptying this field on its own would make every standalone row sort as zero incidents.

api/src/handlers/list_validators.rs
```diff
@@ -521,6 +533,14 @@ pub async fn handler(
...
 ) -> Result<impl Reply, warp::Rejection> {
     metrics::REQUEST_COUNT_VALIDATORS.inc();
+    if let Some(window) = query_params.incident_window_epochs {
+        if window == 0 || window > DEFAULT_CACHE_EPOCHS {
+            return Ok(response_error(
+                StatusCode::BAD_REQUEST,
+                format!("incident_window_epochs must be between 1 and {DEFAULT_CACHE_EPOCHS}"),
+            ));
+        }
+    }
```
ochaloup, line 543: 🤖 claude: Before this change there was no window parameter to reject, so these two conditions are entirely new behaviour and neither is exercised — `grep "must be between" api/src store/src` finds only the `format!` on line 540. The sibling rejection did get covered, but only at the message builder: `incidents_order_rejection` has two unit tests in `api/src/utils/order.rs`, while the branches that call it (`list_providers.rs:57`, `list_clients.rs:55`) do not, so deleting either call site leaves the whole suite green.

What it should be: lift the range check into a small pure `fn validate_incident_window(Option<u64>) -> Result<(), String>` and test `0`, `1`, `DEFAULT_CACHE_EPOCHS` and `DEFAULT_CACHE_EPOCHS + 1` — that is testable without a `WrappedContext`, which is the reason the handler itself cannot be. If the range check is judged too small to be worth extracting, the trade is fine, but then the group-ordering rejection is the one regression that a test can actually catch and the call sites are where it needs catching.

ochaloup (commented):
It seems good at first sight. Just when I work with Claude, it seems there are still some gaps. Not big ones, but possibly good to consider. Can you check my comments, please?

store/src/groups.rs
ilmari-h: didn't see this as insightful enough to include in a comment

api/src/handlers/list_validators.rs
ilmari-h: added this, seemed useful to document

store/src/groups.rs
ilmari-h: seems very nitpicky so ignoring

store/src/utils.rs
ilmari-h: This would've been annoying to debug, nice one. Fixed

store/src/groups.rs
ilmari-h: This was a good point, I gave the `incidents` some more attention now. Turned it into an enum that can be either a vector or a numeric value. For groups where we dont want to list individual incidents (very large groups like clients and providers) incidents is an integer. For validators its a vector as before. Also this now re-enables sorting e.g. /providers by incident count, which led to some code smells commented on by Kron.

api/src/handlers/list_validators.rs
ilmari-h: 🤖 Skipping this one. It's a two-condition range check on a u64; extracting it behind a helper to make it reachable from a test is more indirection than the logic carries.

The other half of the comment no longer applies. incidents_order_rejection and both of its call sites are gone — order_field=Incidents now works on /clients and /providers, since their rows carry GroupIncidents::Count and the column has a real value to sort on rather than an always-empty vector. That removes two of the three untested 400 branches outright, leaving this one.

ochaloup (approved):

PR #268, opened 2026-09-02
[GEN-8870] in /clients fall back to last reported client info

ilmari-h:
`/clients` would return "Unknown" client ID for entries where gossip not yet landed for current epoch.

## Summary by CodeRabbit

- **Improvements**
  - Client grouping now displays labels and lineage information from the latest served-epoch records.
  - Grouping falls back to reported client information when current records are unavailable.
  - Client-tree relationships now use consistent current-record information.
  - Historical grouping continues to use epoch statistics.
  - Grouped and singleton results now show incident details or counts appropriate to each group type.
  - Improved handling of ordering and empty groups.

### DIFF
--- a/store/src/groups.rs
+++ b/store/src/groups.rs
@@ -60,11 +60,6 @@ fn is_unknown_placeholder(value: &str) -> bool {
     }
 }
 
-/// The client the node reported, for one absent from client-ids.csv.
-fn reported_client(client_id_raw: Option<String>) -> Option<String> {
-    normalized(client_id_raw)
-}
-
 /// The registry renders a client lowercase (`agave`); block engine labels are title-cased.
 fn as_client_name(client: String) -> String {
     let mut characters = client.chars();
@@ -86,8 +81,9 @@ fn group_key(
             normalized(operators::operator_of(&validator.vote_account).map(str::to_string))
         }
         GroupKind::ProviderAso => normalized(stats.dc_aso.clone()),
+        // The raw rendering is what the node reported, kept for a client absent from client-ids.csv.
         GroupKind::ClientLabel => normalized(Some(client_label(client_id)))
-            .or_else(|| reported_client(stats.client_id_raw.clone())),
+            .or_else(|| normalized(stats.client_id_raw.clone())),
         GroupKind::ClientLineage => normalized(client_lineage(client_id)).map(as_client_name),
     }
 }
@@ -100,7 +96,7 @@ fn current_group_key(
 ) -> Option<String> {
     match kind {
         GroupKind::ClientLabel => normalized(Some(validator.client_label.clone()))
-            .or_else(|| reported_client(validator.client_id_raw.clone())),
+            .or_else(|| normalized(validator.client_id_raw.clone())),
         GroupKind::ClientLineage => {
             normalized(validator.client_lineage.clone()).map(as_client_name)
         }
--- a/store/src/groups.rs
+++ b/store/src/groups.rs
@@ -57,8 +57,8 @@ fn is_unknown_placeholder(value: &str) -> bool {
 }
 
 /// The client the node reported, for one absent from client-ids.csv.
-fn reported_client(stats: &ValidatorEpochStats) -> Option<String> {
-    normalized(stats.client_id_raw.clone())
+fn reported_client(client_id_raw: Option<String>) -> Option<String> {
+    normalized(client_id_raw)
 }
 
 /// The registry renders a client lowercase (`agave`); block engine labels are title-cased.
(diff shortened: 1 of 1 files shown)


store/src/groups.rs
```diff
@@ -61,8 +61,8 @@ fn is_unknown_placeholder(value: &str) -> bool {
 }
 
 /// The client the node reported, for one absent from client-ids.csv.
-fn reported_client(stats: &ValidatorEpochStats) -> Option<String> {
-    normalized(stats.client_id_raw.clone())
+fn reported_client(client_id_raw: Option<String>) -> Option<String> {
```
kronael: why would the user not call this directly?

store/src/groups.rs
kronael, line 1: Does the user (frontend) know which epoch the record originated at, e.g. how old it is?  e.g. whether its not outdated?

kronael (dismissed):

store/src/groups.rs
ilmari-h: It's the last epoch we got node info for that validator, so it's always the latest we know. If the client changes, the API reports it as soon as we see it.

store/src/groups.rs
ilmari-h: changed to call this directly, yes this was left behind by ai for no reason

store/src/groups.rs
kronael: I know, but what I mean is whether the frontend knows it.  What I meant is that it is a good practice to always include some timing on data that changes.  But I do not understand requirements fully here...

kronael (approved):

store/src/groups.rs
```diff
@@ -424,7 +436,7 @@ fn aggregate_keyed(population: &Population, kind: GroupKind) -> KeyedGroups {
             continue;
         };
 
-        let key = group_key(validator, stats, kind);
+        let key = current_group_key(validator, stats, kind);
```
ochaloup, line 439: 🤖 claude: This keys the current bucket with `current_group_key` while `group_stake_at` still keys the 7d/30d baselines with `group_key`, so the two sides of every stake delta now resolve through different functions. A validator that is gossip-blank both in the current epoch and at the baseline epoch but was observed in between lands under its projected client on one side and under `Unknown` on the other — with a two-member fixture I get Agave reporting `stake_delta_7d = +500` where it was `0` before this change, and giving `group_stake_at` the same newest-row-at-or-below-epoch fallback would make both sides symmetric again.

store/src/groups.rs
```diff
@@ -86,13 +81,30 @@ fn group_key(
...
+    stats: &ValidatorEpochStats,
+    kind: GroupKind,
+) -> Option<String> {
+    match kind {
+        GroupKind::ClientLabel => normalized(Some(validator.client_label.clone()))
+            .or_else(|| normalized(validator.client_id_raw.clone())),
+        GroupKind::ClientLineage => {
+            normalized(validator.client_lineage.clone()).map(as_client_name)
+        }
+        // Neither is projected onto the record, so both read the epoch as stored.
```
ochaloup, line 103: 🤖 claude: nit — `dc_aso` *is* on the record (`dto.rs`, filled in `load_validators` from the newest row), so the comment sends a reader looking for something that is not true; only `Operator` is genuinely absent from the record, since it comes from the operators CSV and has no epoch dimension at all.

```suggestion
        // Only the client columns are gap-filled onto the record; these two read the epoch as stored.
```

store/src/groups.rs
```diff
@@ -647,10 +664,28 @@ mod tests {
...
+                // any of them, so the record keeps a client the epoch being served has not
+                // observed yet. The fixture's raw rendering is the member's, so a row counts as
+                // reporting when it carries either half.
+                let projected_client_id = member
+                    .epochs
+                    .iter()
+                    .find(|(_, _, client_id, _)| {
+                        client_id.is_some() || member.client_id_raw.is_some()
+                    })
+                    .and_then(|(_, _, client_id, _)| *client_id);
```
ochaloup, line 677: 🤖 claude: `member.client_id_raw.is_some()` does not depend on the closure argument, so for any member carrying a raw client this `find` degenerates to "take the newest epoch" and the fallback the fixture exists to model is switched off. Together with the record's `client_id_raw` being taken from the member rather than from the chosen row (line 684), `an_unregistered_client_reported_this_epoch_stays_unclassified` passes identically if the raw had only been reported in `PREVIOUS_EPOCH`, and the case "current epoch blank, an older epoch reported an unregistered readable name" is left untested — moving `client_id_raw` into `EpochSpec` would fix both.

ochaloup (commented):
@ilmari-h adding few comments on this. I think the one `member.client_id_raw` could be considered as a good to fix but feel free to deffer.

PR #269, opened 2026-09-03
[GEN-8776] block production incidents

ilmari-h:
New incident type `BlockProduction`.

The endpoint `GET /validators` now also returns incidents caused by validator failing to meet following criteria:

```
1.  leader_slots  >= 64
2.  missed_slots  >= 4
3.  skip_rate     >= min( max(1%, 10 * cluster_skip_rate[N]), 5% )
```

The API includes query parameters that allow caller to adjust leader slots and missed slots thresholds (`min_incident_leader_slots` and `min_incident_missed_slots`). 64 and 4 respectively are minimums; events with lower values are not recorded as incidents.

When `BlockProduction` is caused by `Downtime` incident, these are combined into 1 incident record instead of 2.

All `Downtime` incidents now include also block production details if available:

```rs
pub struct BlockProductionDetail {
    pub leader_slots: u64,
    pub blocks_produced: u64,
    pub missed_slots: u64,
    /// `missed_slots / leader_slots`, as a fraction.
    pub skip_rate: f64,
    /// The epoch's cluster-wide skip rate, over the validators that were evaluable that epoch.
    pub cluster_skip_rate: f64,
    /// The bar `skip_rate` had to clear that epoch, as a fraction.
    pub threshold: f64,
    /// Whether these numbers count as an incident.
    /// Affected by query params `min_incident_missed_slots` and `min_incident_leader_slots`.
    pub counts_as_incident: bool,
}
```
This struct is included also in the lone `BlockProduction` incident record. This includes information that allows the client to render more details about why the downtime/block production incident happened.

## Summary by CodeRabbit

- **New Features**
  - Validator listings now support filtering by incident type, including downtime and block-production incidents.
  - Added configurable missed-slot and leader-slot thresholds for identifying block-production incidents.
  - Incident details now distinguish downtime from block-production events and provide clearer timing and classification information.
  - Added incident detail types to the API documentation schema.

- **Bug Fixes**
  - Prevented incidents from being created from incomplete, still-running epochs.
  - Improved incident filtering, grouping, deduplication, and mixed-condition handling.
  - Corrected incident ordering to use the block-production period start time.

### DIFF
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -457,7 +457,6 @@ pub fn filter_validators(
                 return false;
             }
             match &incident.detail {
-                // Restart noise is under the floor, unless the epoch also breached block production.
                 IncidentDetail::Downtime {
                     downtime_seconds,
                     block_production,
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -476,7 +476,7 @@ pub fn filter_validators(
     let min_incident_downtime = config
         .min_incident_downtime_seconds
         .unwrap_or(DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS);
-    // `breached` owns both defaults, so the caller's floors travel as they arrived.
+    // `counts_as_incident` owns both defaults, so the caller's floors travel as they arrived.
     let (min_missed_slots, min_leader_slots) = (
         config.min_incident_missed_slots,
         config.min_incident_leader_slots,
@@ -498,26 +498,29 @@ pub fn filter_validators(
             .unwrap_or(DEFAULT_INCIDENTS_WINDOW_EPOCHS),
     );
     for validator in validators.values_mut() {
-        validator.incidents.retain(|incident| {
+        validator.incidents.retain_mut(|incident| {
             if incident.epoch < from_epoch {
                 return false;
             }
-            // Served when any symptom is asked for and over its own floor.
-            let (downtime_seconds, block_production) = match &incident.detail {
+            let (downtime_seconds, block_production) = match &mut incident.detail {
                 IncidentDetail::Downtime {
                     downtime_seconds,
                     block_production,
                     ..
-                } => (Some(*downtime_seconds), block_production.as_ref()),
+                } => (Some(*downtime_seconds), block_production.as_mut()),
                 IncidentDetail::BlockProduction {
                     block_production, ..
                 } => (None, Some(block_production)),
             };
+            // Re-answered against the caller's floors, so what is served agrees with what is said.
+            let counts_as_incident = block_production.is_some_and(|detail| {
+                detail.counts_as_incident = detail.counts_as_incident(min_missed_slots, min_leader_slots);
+                detail.counts_as_incident
+            });
+            // Served when any symptom is asked for and over its own floor.
             (wants_downtime
                 && downtime_seconds.is_some_and(|seconds| seconds >= min_incident_downtime))
-                || (wants_block_production
-                    && block_production
-                        .is_some_and(|detail| detail.breached(min_missed_slots, min_leader_slots)))
+                || (wants_block_production && counts_as_incident)
         });
     }
 
@@ -995,14 +998,17 @@ mod tests {
     const FIXTURE_MISSED_SLOTS: u64 = 548;
 
     fn skipped(missed_slots: u64) -> BlockProductionDetail {
-        BlockProductionDetail {
+        let mut detail = BlockProductionDetail {
             leader_slots: FIXTURE_LEADER_SLOTS,
             blocks_produced: FIXTURE_LEADER_SLOTS - missed_slots,
             missed_slots,
             skip_rate: missed_slots as f64 / FIXTURE_LEADER_SLOTS as f64,
             cluster_skip_rate: 0.001_57,
             threshold: 0.015_7,
-        }
+            counts_as_incident: false,
+        };
+        detail.counts_as_incident = detail.counts_as_incident(None, None);
+        detail
     }
 
     fn block_production_incident_of(epoch: u64, missed_slots: u64) -> IncidentRecord {
@@ -1295,7 +1301,7 @@ mod tests {
     }
 
     #[test]
-    fn a_downtime_carrying_numbers_that_did_not_breach_is_no_block_production_incident() {
+    fn a_downtime_carrying_numbers_under_the_rule_is_no_block_production_incident() {
         // 5 of 6392 is 0.08%, under the bar: the numbers are information, not a verdict.
         let validators = map(vec![with_incidents("blip", vec![down_and_skipped(600, 5)])]);
         let config = GetValidatorsConfig {
@@ -1308,7 +1314,7 @@ mod tests {
     }
 
     #[test]
-    fn a_downtime_carrying_numbers_that_breached_is_a_block_production_incident() {
+    fn a_downtime_carrying_numbers_over_the_rule_is_a_block_production_incident() {
         let validators = map(vec![with_incidents(
             "skipper",
             vec![down_and_skipped(600, FIXTURE_MISSED_SLOTS)],
@@ -1318,6 +1324,29 @@ mod tests {
             ..config()
         };
         assert_eq!(filter_validators(validators, &config)[0].incidents.len(), 1);
+    }
+
+    // What is served has to agree with what it says about itself.
+    #[test]
+    fn a_caller_floor_rewrites_the_counts_as_incident_flag_it_is_served_with() {
+        let validators = map(vec![with_incidents(
+            "skipper",
+            vec![down_and_skipped(600, FIXTURE_MISSED_SLOTS)],
+        )]);
+        let config = GetValidatorsConfig {
+            min_incident_missed_slots: Some(FIXTURE_MISSED_SLOTS + 1),
+            ..config()
+        };
+
+        let served = filter_validators(validators, &config);
+        let IncidentDetail::Downtime {
+            block_production, ..
+        } = &served[0].incidents[0].detail
+        else {
+            panic!("a downtime fixture is a downtime incident");
+        };
+        // Kept by its downtime, but no longer a block production incident under this floor.
+        assert!(!block_production.as_ref().unwrap().counts_as_incident);
     }
 
     #[test]
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -915,7 +915,6 @@ mod tests {
                     skip_rate: 0.0857,
                     cluster_skip_rate: 0.001_57,
                     threshold: 0.015_7,
-                    cluster_skip_rate_multiple: Some(54.6),
                 },
             },
         }
--- a/api/src/handlers/list_validators.rs
+++ b/api/src/handlers/list_validators.rs
@@ -17,14 +17,13 @@ use store::{
         ValidatorsAggregated,
     },
     groups::{aggregate_operators, singleton_group},
-    incidents::MIN_MISSED_SLOTS,
+    incidents::{MIN_LEADER_SLOTS, MIN_MISSED_SLOTS},
     utils::{to_fixed_for_sort, worst_known_commission, DEFAULT_CACHE_EPOCHS},
 };
 use warp::{http::StatusCode, reply::json, Reply};
 
 const DEFAULT_EPOCHS: usize = 15;
 const DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS: u64 = 180;
-const DEFAULT_MIN_INCIDENT_MISSED_SLOTS: u64 = MIN_MISSED_SLOTS;
 const DEFAULT_INCIDENTS_WINDOW_EPOCHS: u64 = 90;
 const DEFAULT_LIMIT: usize = 100;
 
@@ -103,6 +102,8 @@ pub struct QueryParams {
     min_incident_downtime_seconds: Option<u64>,
     /// Minimum missed leader slots for a skipped epoch to read as an incident. Defaults to 4, minimum 4.
     min_incident_missed_slots: Option<u64>,
+    /// Minimum leader slots an epoch needs before its block production is judged. Defaults to 64, minimum 64.
+    min_incident_leader_slots: Option<u64>,
     /// Epochs back the `incidents` array reaches, counting the newest reported epoch itself. Defaults to 90; above 90 — the whole window the cache holds — answers 400. Unrelated to `epochs`, which sizes `epoch_stats`.
     incident_window_epochs: Option<u64>,
     query_verified: Option<bool>,
@@ -135,6 +136,7 @@ pub struct GetValidatorsConfig {
     pub query_incident_types: Option<Vec<IncidentType>>,
     pub min_incident_downtime_seconds: Option<u64>,
     pub min_incident_missed_slots: Option<u64>,
+    pub min_incident_leader_slots: Option<u64>,
     pub incident_window_epochs: Option<u64>,
     pub query_verified: Option<bool>,
     pub query_protected: Option<bool>,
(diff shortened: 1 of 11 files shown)


store/src/utils.rs
```diff
@@ -254,21 +255,90 @@ pub async fn load_incidents(
...
+    // the validator it judges come from one snapshot.
+    let cluster_skip_rates = crate::incidents::cluster_skip_rates(records.values());
+
+    for (vote_account, record) in records {
+        // Same window the query above read.
+        for stats in record
+            .epoch_stats
+            .iter()
+            .filter(|stats| (from_epoch..=last_epoch).contains(&stats.epoch))
+        {
```
ochaloup, line 284: 🤖 claude: `last_epoch` is `MAX(epoch) FROM validators`, so this range includes the epoch in flight — and for the current epoch `get_block_production_by_validator` returns only the elapsed part, so a validator that misses one leader turn inside its first 64 slots reads 6.25%, gets flagged, and is un-flagged hours later once it settles at 0.1%. `query_incident_free` and `order_field=incidents` flap with it. The ticket's analysis deliberately excluded the in-progress epoch; gating the standalone incident below on `stats.epoch_end_at` being `Some` would match that.

store/src/utils.rs
```diff
@@ -254,21 +255,90 @@ pub async fn load_incidents(
...
+                    ..
+                } = &mut incident.detail
+                {
+                    *downtime_block_production = Some(block_production.clone());
+                    carried = true;
+                }
+            }
+
+            // With no downtime record to attach to, check block production rule and classify as incident if counts_as_incident.
+            if !carried && block_production.counts_as_incident {
```
ochaloup: 🤖 claude: nit — the comment restates the condition beneath it, and the rule was already evaluated in `BlockProductionDetail::for_epoch`; this line only reads the answer.

```suggestion
            if !carried && block_production.counts_as_incident {
```

store/src/utils.rs
```diff
@@ -254,21 +255,90 @@ pub async fn load_incidents(
...
+                // Without the `epochs` row there is nothing to anchor the incident to.
+                let Some(epoch_start_at) = stats.epoch_start_at else {
+                    continue;
+                };
```
ochaloup: 🤖 claude: This guard never fires — `load_validators` substitutes `Utc::now()` whenever the `epochs` join yields NULL, so `epoch_start_at` is already `Some` by the time it arrives here. The effect is worse than dead code: an open-epoch incident is then served with `epoch_start_at` set to the cache-refresh clock instead of the epoch's real start, minutes old rather than days.

store/src/dto.rs
```diff
@@ -386,13 +384,64 @@ pub struct UptimeRecord {
...
+}
+
+impl IncidentDetail {
+    /// When the incident started, for ordering: a downtime interval when it went down, a block
+    /// production epoch when that epoch ended. `None` for an epoch with no end on record, which is
+    /// the running one.
+    pub fn started_at(&self) -> Option<DateTime<Utc>> {
+        match self {
+            Self::Downtime { start_at, .. } => Some(*start_at),
+            Self::BlockProduction { epoch_end_at, .. } => *epoch_end_at,
```
ochaloup: 🤖 claude: `epoch_end_at` is `None` for the open epoch and `None` sorts before every `Some`, so `GroupIncidents::sort` puts an incident from the epoch in flight ahead of every older one — I reproduced epoch order `[101, 10, 100]` where `[10, 100, 101]` was expected, on `operators[].incidents`. Reading `epoch_start_at` fixes the ordering and makes the method name accurate at the same time.

```suggestion
            Self::BlockProduction { epoch_start_at, .. } => Some(*epoch_start_at),
```

store/src/dto.rs
```diff
@@ -386,13 +384,64 @@ pub struct UptimeRecord {
...
     pub epoch: u64,
-    pub start_at: DateTime<Utc>,
-    pub end_at: DateTime<Utc>,
-    pub downtime_seconds: u64,
+    #[serde(flatten)]
+    pub detail: IncidentDetail,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, utoipa::ToSchema)]
+#[serde(tag = "incident_type")]
```
ochaloup, line 397: 🤖 claude: The tag serialises PascalCase — `"incident_type":"BlockProduction"` on the wire — while `query_incident_types` takes `block_production`, so the value a client reads back is not the value it sends, and the `"blockproduction"` arm in `parse_list` exists only to paper over that. Both sides are new in this PR, so aligning them is free now and breaking once anything consumes the field.

```suggestion
#[serde(tag = "incident_type", rename_all = "snake_case")]
```

store/src/dto.rs
```diff
@@ -386,13 +384,64 @@ pub struct UptimeRecord {
...
+/// One incident on a validator in one epoch. `incident_type` says which kind, and which of the
+/// remaining fields are present.
 #[derive(Deserialize, Serialize, Debug, Clone, utoipa::ToSchema)]
 pub struct IncidentRecord {
     pub epoch: u64,
-    pub start_at: DateTime<Utc>,
-    pub end_at: DateTime<Utc>,
-    pub downtime_seconds: u64,
+    #[serde(flatten)]
+    pub detail: IncidentDetail,
```
ochaloup, line 393: 🤖 claude: This `flatten` is the whole reason `start_at` / `end_at` / `downtime_seconds` stay top-level for existing downtime consumers, and nothing asserts it — drop the attribute in a later refactor and the payload silently nests under `detail`, breaking every one of them while the suite stays green. One test on `serde_json::to_value` of each variant would pin the contract.

store/src/dto.rs
```diff
@@ -386,13 +384,64 @@ pub struct UptimeRecord {
...
+pub struct BlockProductionDetail {
+    pub leader_slots: u64,
+    pub blocks_produced: u64,
+    pub missed_slots: u64,
+    /// `missed_slots / leader_slots`, as a fraction.
+    pub skip_rate: f64,
+    /// The epoch's cluster-wide skip rate, over the validators that were evaluable that epoch.
+    pub cluster_skip_rate: f64,
+    /// The bar `skip_rate` had to clear that epoch, as a fraction.
+    pub threshold: f64,
```
ochaloup, line 428: 🤖 claude: The ticket lists "the resulting multiple" among the fields the reason string needs — "55x more than baseline" — and `threshold` is served in its place; a client deriving it as `skip_rate / cluster_skip_rate` gets a division by zero whenever the cluster skipped nothing, and JSON has no `Infinity`. Worth settling whether the FE derives it or expects it served.

store/src/incidents.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+        let (leader_slots, blocks_produced) = totals.entry(stats.epoch).or_default();
+        *leader_slots += stats.leader_slots;
+        // A node reporting more blocks than slots is upstream noise; it must not mint slots here.
+        *blocks_produced += stats.blocks_produced.min(stats.leader_slots);
+    }
+
+    totals
+        .into_iter()
+        .filter(|(_, (leader_slots, _))| *leader_slots > 0)
```
ochaloup: 🤖 claude: nit — unreachable: an entry is only created after the `stats.leader_slots < MIN_LEADER_SLOTS` continue above, so every accumulated `leader_slots` is at least 64.

```suggestion
        .into_iter()
```

api/src/handlers/list_validators.rs
```diff
@@ -23,6 +27,27 @@ const DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS: u64 = 180;
 const DEFAULT_INCIDENTS_WINDOW_EPOCHS: u64 = 90;
 const DEFAULT_LIMIT: usize = 100;
 
+/// Which kind of incident a caller wants served.
+#[derive(Deserialize, Serialize, Debug, Clone, Copy, PartialEq, Eq)]
```
ochaloup: 🤖 claude: nit — `Deserialize` and `Serialize` are unused: the query string is parsed by hand in `parse_list` and `GetValidatorsConfig` derives only `Debug`, so this type never reaches serde. `Eq` is unused too — `types.contains` needs `PartialEq` alone.

```suggestion
#[derive(Debug, Clone, Copy, PartialEq)]
```

api/src/handlers/list_validators.rs
```diff
@@ -69,10 +94,16 @@ pub struct QueryParams {
     query_marinade_stake: Option<bool>,
     query_with_names: Option<bool>,
     query_sfdp: Option<bool>,
-    /// `true` keeps the validators whose `incidents` array comes back empty, `false` the rest. It reads that array, so `min_incident_downtime_seconds` and `incident_window_epochs` shape it too, where `epochs` and `query_from_date` do not.
+    /// `true` keeps the validators whose `incidents` array comes back empty, `false` the rest. It reads that array, so `query_incident_types`, `min_incident_downtime_seconds`, `min_incident_missed_slots` and `incident_window_epochs` shape it too, where `epochs` and `query_from_date` do not.
```
ochaloup: 🤖 claude: nit — `min_incident_leader_slots` shapes this array just as much, as `a_leader_slot_floor_above_the_default_drops_what_the_rule_admitted` shows, but it is missing from the list.

```suggestion
    /// `true` keeps the validators whose `incidents` array comes back empty, `false` the rest. It reads that array, so `query_incident_types`, `min_incident_downtime_seconds`, `min_incident_missed_slots`, `min_incident_leader_slots` and `incident_window_epochs` shape it too, where `epochs` and `query_from_date` do not.
```

api/src/handlers/list_validators.rs
```diff
@@ -442,15 +476,52 @@ pub fn filter_validators(
     let min_incident_downtime = config
         .min_incident_downtime_seconds
         .unwrap_or(DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS);
+    // `counts_as_incident` owns both defaults, so the caller's floors travel as they arrived.
+    let (min_missed_slots, min_leader_slots) = (
+        config.min_incident_missed_slots,
+        config.min_incident_leader_slots,
+    );
```
ochaloup: 🤖 claude: nit — this only renames two `Copy` `Option<u64>` fields that are each read once, in the `counts_as_incident` call below; passing `config.min_incident_missed_slots` and `config.min_incident_leader_slots` straight in says the same thing. Keep the comment where it sits — it earns its place explaining the absent `unwrap_or` beside `min_incident_downtime`'s.

api/src/handlers/list_validators.rs
```diff
@@ -442,15 +476,52 @@ pub fn filter_validators(
...
+            // Re-answered against the caller's floors, so what is served agrees with what is said.
+            let counts_as_incident = block_production.is_some_and(|detail| {
+                detail.counts_as_incident =
+                    detail.counts_as_incident(min_missed_slots, min_leader_slots);
+                detail.counts_as_incident
+            });
+            // Served when any symptom is asked for and over its own floor.
+            (wants_downtime
+                && downtime_seconds.is_some_and(|seconds| seconds >= min_incident_downtime))
+                || (wants_block_production && counts_as_incident)
```
ochaloup: 🤖 claude: A validator with three DOWN intervals in one skipped epoch answers `query_incident_types=block_production` with three rows, each carrying `"incident_type":"Downtime"` and the identical `block_production` object — so a caller that asked for block production gets rows typed Downtime, and one counting them counts a single event three times. That is the "one event with two symptoms, not two rows" criterion the store honours and this projection undoes; `incidents` is already epoch-sorted upstream, so keeping only the first row that the block-production clause alone kept is enough.

store/tests/incidents_sql.rs
```diff
@@ -209,26 +220,139 @@ async fn incidents_are_keyed_by_the_validator_that_was_down() {
         "2026-01-01T00:05:00Z",
     )
     .await;
+
+    let incidents = load_incidents(&client, 100, 100, &skipped_epoch("voteA", 100))
+        .await
+        .unwrap();
+
+    assert_eq!(epochs(&incidents["voteA"]), vec![100]);
```
ochaloup: 🤖 claude: `load_incidents` now has two independent paths into the same map — the SQL loop keyed by the row's `vote_account` and the new per-validator loop keyed by the `records` key — and `incidents_are_keyed_by_the_validator_that_was_down` was dropped, so nothing covers the keying any more. A mix-up would attach one validator's block production to another's downtime silently; a second validator here that only skipped would cover the keying and the enrichment/standalone split in one assertion.

ochaloup (commented):
I was thinking it's good but Claude raised quite a big number of findings that seem really relevant. Sorry if that's overhelming but all of them seems to me good to check.

store/src/utils.rs
ilmari-h: fixed

store/src/utils.rs
ilmari-h: > The effect is worse than dead code: an open-epoch incident is then served with epoch_start_at set to the cache-refresh clock instead of the epoch's real start, minutes old rather than days.

🤖 Claim wrong, epoch_start_at served: was true, now moot. epochs.start_at and epochs.end_at are both NOT NULL (migrations/0000-init.sql:104-105) and come from the same LEFT JOIN epochs ON epochs.epoch = validators.epoch (utils.rs:1058). They are null together — only when the epochs row is missing. So gating on epoch_end_at being Some excludes exactly the rows whose epoch_start_at was the Utc::now() substitute. Any incident built now carries the real epoch start.

store/src/dto.rs
ilmari-h: fixed

store/src/dto.rs
ilmari-h: made this consistent with rest of API contract, using PascalCase now consistently

store/src/dto.rs
ilmari-h: doesnt warrant test imo

store/src/dto.rs
ilmari-h: leaving this up to frontend to calculate on purpose

store/src/incidents.rs
ilmari-h: fixed

api/src/handlers/list_validators.rs
ilmari-h: simplified this comment

api/src/handlers/list_validators.rs
ilmari-h: this got me thinking, I will rewrite the incident grouping better so we can have the caller dictate the limits, what counts as incident and how they're grouped

ilmari-h: rewrote this: cache now stores incidents without merging them. Incidents are merged/filtered at request time depending on what caller requested.

store/tests/incidents_sql.rs
ilmari-h: fixed

store/src/utils.rs
```diff
@@ -254,21 +255,92 @@ pub async fn load_incidents(
...
+                    carried = true;
+                }
+            }
+
+            if !carried && block_production.counts_as_incident {
+                // No `end_at` means the epoch is still running, and its block production covers
+                // just the leader slots so far: skip to not produce misleading/false incident
+                let (Some(epoch_start_at), Some(epoch_end_at)) =
+                    (stats.epoch_start_at, stats.epoch_end_at)
+                else {
```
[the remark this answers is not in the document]
ilmari-h: fixed in recent refactor

api/src/handlers/list_validators.rs
ilmari-h: fixed

ilmari-h:
@ochaloup thanks for the review, fixed most of these points and refactored how we store the incidents in cache. Asking for another round

ochaloup (dismissed):

ilmari-h:
@ochaloup spotted one more issue, bw compatability with frontend. Added 1 more commit, asking for re-approval

ochaloup (approved):

PR #271, opened 2026-09-07
fix(api): add the missing leading slash to two documented report paths

dualisimo:
## Summary

`reports/scoring` and `reports/commission-changes` are annotated in `utoipa::path` without a leading slash, so they land in `/docs.json` as `"reports/scoring"` instead of `"/reports/scoring"`. Strict OpenAPI validators reject that (`#/paths must NOT have unevaluated properties`), which breaks client generators reading the spec — orval 8 now fails hard on it, where orval 7 only warned and dropped the two operations.

Routing is unaffected: `main.rs` registers both under `warp::path!("reports" / ...)`, and both endpoints already return 200 in production. Only the published spec changes. The sibling handler `reports_staking.rs` was already correct; these were the only two occurrences in the repo.

Added a test over `ApiDoc::openapi()` asserting every documented path starts with a slash — it fails on master with `paths missing leading slash: ["reports/commission-changes", "reports/scoring"]` and passes with the fix.

## Follow-ups

- None.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the commission-changes reporting endpoint path to `/reports/commission-changes`.
  - Corrected the scoring reports endpoint path to `/reports/scoring`.
  - These corrections ensure reporting API requests use consistent, valid paths.

- **Tests**
  - Added validation to confirm that all generated API documentation paths begin with `/`.

### DIFF
--- a/api/src/api_docs.rs
+++ b/api/src/api_docs.rs
@@ -115,3 +115,28 @@ use utoipa::OpenApi;
     )
 )]
 pub struct ApiDoc;
+
+#[cfg(test)]
+mod tests {
+    use super::ApiDoc;
+    use utoipa::OpenApi;
+
+    /// A `utoipa::path` without a leading slash still routes correctly but
+    /// produces a spec that strict OpenAPI validators reject, which breaks
+    /// client generators consuming /docs.json.
+    #[test]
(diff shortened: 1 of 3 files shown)


ilmari-h (approved):
nice one

PR #273, opened 2026-09-10
[GEN-8811] serve information about releases and version floors

ilmari-h:
Added `GET /releases`

This data will be used as basis for the "late client version" incident type

Example response:
```
{
  "releases": [
    {
      "available_epoch": 1016,
      "client_lineage": "agave",
      "client_version": "4.3.0-beta.0",
      "release_url": "https://github.com/anza-xyz/agave/releases/tag/v4.3.0-beta.0",
      "released_at": "2026-08-14T18:34:44Z",
      "updated_at": "2026-09-10T07:11:50.467767Z"
    }
  ],
  "feature_gate_floors": [
    {
      "announced_epoch": 1015,
      "client_lineage": "agave",
      "client_version": "4.2.0-beta.1",
      "effective_epoch": 1019
    }
  ],
  "sfdp_floors": [
    {
      "client_lineage": "agave",
      "client_version": "4.2.2",
      "effective_epoch": 1033
    }
  ]
}
```

We get the SFDP floors from the [Solana API](https://api.solana.org/api/community/v1/sfdp_required_versions?cluster=mainnet-beta) and newest release info from the GitHub releases of [anza-xyz/agave](https://github.com/anza-xyz/agave/releases) and [firedancer-io/firedancer](https://github.com/firedancer-io/firedancer/releases), collected by a collector service.

Feature gate floors are derived: [Anza's tracker](https://raw.githubusercontent.com/wiki/anza-xyz/agave/feature-gate-tracker-schedule.json) gives the version each gate shipped in, the gate's own account on chain gives the epoch it activated, and the floor is the running maximum of the two. [`version-floor.json`](https://raw.githubusercontent.com/wiki/anza-xyz/agave/version-floor.json) bounds the read set and cross-checks the result. History before that is seeded by the migration, and a derived floor overwrites a seeded epoch — the tracker and the chain are the source of truth.

One row per `(client_lineage, client_version)`, with the fetchers writing disjoint columns, so there is no `source` column and no precedence between sources.

Also went and wrote a common `csv` module. We now have 4 CSV files in this repo that all implement their own parser. Moved that to 1 module.

## Summary by CodeRabbit

- **New Features**
  - Added release tracking for client versions, publication details, availability epochs, SFDP floors, and feature-gate floors.
  - Added a `/releases` API with client, epoch, and sort-order filters.
  - Added commands and scripts to collect release data and store it in PostgreSQL.
  - Added support for collecting releases from GitHub and SFDP sources.
- **Documentation**
  - Documented the releases API, collection workflow, and storage commands.
- **Bug Fixes**
  - Improved CSV validation and error reporting for registry and configuration data.


collect/src/releases/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Gossip reports `26.8.2` where the Firedancer tag zero-pads to `v26.08.2`, and no tag carries a
+/// leading zero that means anything, so the padding is dropped to keep one spelling per version.
+pub fn normalize_version(tag: &str) -> String {
+    let version = tag.trim().trim_start_matches('v');
+    let parts: Vec<&str> = version.split('.').collect();
+    if parts.len() == 3
+        && parts
+            .iter()
+            .all(|p| !p.is_empty() && p.bytes().all(|b| b.is_ascii_digit()))
```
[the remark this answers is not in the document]
ilmari-h: done

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
...
+            continue;
+        }
+        previous = Some(version);
+        if let Some((_, first_epoch)) = effective.iter().find(|(v, _)| v == version) {
+            // A floor that was lowered and later raised back: "the floor at epoch E" resolves
+            // against the first run, which is wrong for the second one.
+            warn!(
+                "Floor {version} is effective again at epoch {epoch} after being effective at {first_epoch}"
+            );
+            continue;
```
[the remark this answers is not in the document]
ilmari-h: done, we now override instead of skipping

scripts/collect-store-releases.bash
```diff
@@ -0,0 +1,21 @@
...
+  echo "Collecting releases"
+  # Pass-through args (e.g. --from-epoch 688 for the full floor history, or --source github).
+  ${SCRIPT_DIR}/collect-releases.bash "$@" > "$output_file"
+  if [[ -s "$output_file" ]]; then
+    echo "Storing releases from $output_file"
+    ${SCRIPT_DIR}/store-releases.bash "$output_file"
+  else
+    echo "No release data collected (upstream unreachable); skipping store, will retry next run"
+  fi
+  set +e
```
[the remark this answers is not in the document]
ilmari-h: Same pattern as other scripts, low impact, skipping

store/src/releases.rs
```diff
@@ -0,0 +1,443 @@
...
+        .values()
+        .filter(|row| row.released_at.is_some() && row.available_epoch.is_none())
+        .count();
+    if unplaced > 0 {
+        info!("{unplaced} releases predate the epochs we hold, so they carry no available_epoch");
+    }
+
+    let upserted_availability =
+        upsert_availability(psql_client, &availability, snapshot_created_at).await?;
+    let upserted_floors = upsert_floors(psql_client, &floors, snapshot_created_at).await?;
```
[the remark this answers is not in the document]
ilmari-h: Real, benign, and fixing it would break a convention the repo states out loud.

Every write is an idempotent upsert keyed on (client_lineage, client_version), and nothing ever deletes or invalidates a row. So a half-applied snapshot leaves rows that are individually correct and just missing some updates — the next run writes them. 
wontfix

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
...
+/// has to come from where the value changes, not from that flag.
+fn collapse_runs(series: &BTreeMap<u64, String>) -> Vec<(String, u64)> {
+    let mut effective: Vec<(String, u64)> = Vec::new();
+    let mut previous: Option<&str> = None;
+
+    for (epoch, version) in series {
+        if previous == Some(version.as_str()) {
+            continue;
+        }
+        previous = Some(version);
```
ochaloup: 🤖 claude: `previous` starts as `None` and the series spans only `from_epoch..=to_epoch`, so a floor already in force when the window opens is recorded as taking effect at the window's first epoch — and `upsert_floors` writes that over the backfilled value with no guard. With the default `--epochs-back 20`, agave 4.0.2's correct floor epoch 992 becomes 995, then 996, and freezes at the last epoch of its own run, after which `get_sfdp_floor_at_epoch` answers the predecessor version for every epoch in that run.

Anchoring the fetch on `from_epoch - 1` and dropping a run that starts at the window edge fixes it. `LEAST()` in the upsert would also work but silently swallows the deliberate reinstatement rule just below.

store/src/releases.rs
```diff
@@ -0,0 +1,443 @@
...
+        {
+            return Some(*epoch);
+        }
+
+        // Past the last closed epoch: the running one is the only epoch it can belong to, and it
+        // has no `epochs` row until `store close-epoch` writes one.
+        let (_, _, last_end_at) = self.closed.last()?;
+        if at >= *last_end_at {
+            return self.running;
+        }
```
ochaloup: 🤖 claude: This never checks that `running` immediately follows the last closed epoch, so with `epochs` holding up to 1000 while `cluster_info` already reports 1003, a release published during epoch 1001 is stored as `available_epoch = 1003`. That is precisely the invented lateness the doc-comment above forbids, in the column that exists to measure lateness.

store/src/releases.rs
```diff
@@ -0,0 +1,443 @@
...
+                $2::TEXT[],
+                $3::NUMERIC[],
+                $4::TIMESTAMP WITH TIME ZONE[],
+                $5::TEXT[],
+                $6::TIMESTAMP WITH TIME ZONE[],
+                $7::TIMESTAMP WITH TIME ZONE[]
+            )
+            ON CONFLICT (client_lineage, client_version)
+            DO UPDATE SET
+                available_epoch = EXCLUDED.available_epoch,
```
ochaloup: 🤖 claude: GitHub's full release history is re-fetched and `available_epoch` recomputed on every run, so a single run where `epoch_at` returns `None` — no `cluster_info` row past `MAX(epochs.epoch)` — puts NULL over a value a healthy run computed correctly, for exactly the newest releases. The comment above guards the floor column, not this one.

```suggestion
                available_epoch = COALESCE(EXCLUDED.available_epoch, releases.available_epoch),
```

store/src/releases.rs
```diff
@@ -0,0 +1,443 @@
...
+        .query(
+            &format!(
+                "
+        SELECT DISTINCT ON (client_lineage)
+            client_lineage, client_version, sfdp_floor_epoch
+        FROM {RELEASES_TABLE}
+        WHERE sfdp_floor_epoch IS NOT NULL
+          AND sfdp_floor_epoch <= $1::NUMERIC
+          AND ($2::TEXT IS NULL OR client_lineage = $2::TEXT)
+        ORDER BY client_lineage, sfdp_floor_epoch DESC
```
ochaloup: 🤖 claude: `DISTINCT ON (client_lineage)` with no tiebreaker past the epoch resolves arbitrarily when two rows share `sfdp_floor_epoch` — reachable because floors are pre-recorded for unstarted epochs (`current_epoch + UPCOMING_EPOCHS`) and a retracted one is never cleared. If SFDP revises an upcoming floor, both versions claim the same effective epoch and the retracted one can win.

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
...
+        if let Some((_, effective_epoch)) = effective.iter_mut().find(|(v, _)| v == version) {
+            warn!(
+                "Floor {version} is effective again at epoch {epoch}, replacing epoch {effective_epoch}"
+            );
+            *effective_epoch = *epoch;
+            continue;
+        }
```
ochaloup, line 112: 🤖 claude: One row per version with a single epoch column cannot hold two disjoint runs, so this discards the earlier one. With the module's own test series — 4.1.0-rc.1 at 1000, 4.2.0-rc.1 at 1001, 4.1.0-rc.1 at 1002 — `get_sfdp_floor_at_epoch("agave", 1000)` finds no row `<= 1000` and reports that there was no floor at epoch 1000, though 4.1.0-rc.1 was required then; the test only asserts the later interval.

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
...
+            )?;
+            thread::sleep(REQUEST_INTERVAL);
+            let Some(row) = row else { continue };
+
+            for (versions, min_version) in [
+                (&mut agave, row.agave_min_version),
+                (&mut firedancer, row.firedancer_min_version),
+            ] {
+                if let Some(version) = min_version.filter(|v| !v.trim().is_empty()) {
+                    versions.insert(epoch, version.trim().to_string());
```
ochaloup: 🤖 claude: The SFDP side stores the version raw while `github.rs` runs every tag through `normalize_version` and gates it on `is_plausible_node_version`, and `UNIQUE(client_lineage, client_version)` is exactly what joins "when it shipped" to "when it became required" — so a `26.08.2` vs `26.8.2` spelling difference splits one version into two rows with no error anywhere. A leading `v` would additionally make `firedancer_lineage` file a Frankendancer floor under `firedancer`.

```suggestion
                    versions.insert(epoch, super::normalize_version(version.trim()));
```

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
...
+    to_epoch: u64,
+    client: reqwest::blocking::Client,
+}
+
+impl SfdpFetcher {
+    pub fn new(from_epoch: u64, to_epoch: u64) -> anyhow::Result<Self> {
+        Ok(Self {
+            from_epoch,
+            to_epoch,
+            client: reqwest::blocking::Client::new(),
```
ochaloup: 🤖 claude: `reqwest` applies no default request or connect timeout, so a half-open connection to `api.solana.org` blocks the thread indefinitely — `retry_blocking` never receives an error, and the five-step `FETCH_BACKOFF` added for this endpoint's 429s never engages. A hang is the one failure that ladder cannot see, and `github.rs`'s builder has the same gap.

```suggestion
            client: reqwest::blocking::Client::builder()
                .timeout(Duration::from_secs(30))
                .connect_timeout(Duration::from_secs(10))
                .build()?,
```

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    #[arg(
+        long = "from-epoch",
+        help = "Query version floors from this epoch onwards. Overrides --epochs-back (use for historical backfill; the SFDP API answers from epoch 688 on)."
+    )]
+    from_epoch: Option<u64>,
+
+    #[arg(
+        long = "github-token",
+        env = "GITHUB_TOKEN",
```
ochaloup, line 131: 🤖 claude: clap prints an `env`-backed arg's current value in `--help` unless `hide_env_values` is set. Verified: `GITHUB_TOKEN=ghp_x ./collect releases --help` emits `[env: GITHUB_TOKEN=ghp_x]`, so any `--help` run in a job that has the token in scope writes the credential into the build log.

```suggestion
        env = "GITHUB_TOKEN",
        hide_env_values = true,
```

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ReleasesSnapshot {
+    pub version: u16,
+    pub created_at: String,
+    pub releases: Vec<ReleaseEntry>,
+}
+
+pub trait ReleaseFetcher {
+    fn source(&self) -> ReleaseSource;
```
ochaloup: 🤖 claude: `source()` is implemented by both fetchers and never called — `collect_releases_info` only calls `fetch()`. Three more new public items are in the same state: `ReleaseSource::parse` (only its own round-trip test), `feature_gates::floor_at_epoch` and `releases::get_sfdp_floor_at_epoch` (only `store/tests/releases_sql.rs`); clippy misses all four because they are `pub` in lib crates.

incident-free-validators.json
ochaloup, line 1: I believe this should not be part of the PR.

store/src/releases.rs
```diff
@@ -0,0 +1,443 @@
...
+            &format!(
+                "
+        SELECT
+            client_lineage, client_version, available_epoch, released_at, release_url, updated_at
+        FROM {RELEASES_TABLE}
+        -- A row with no timestamp exists only because SFDP named the version as a floor; it belongs
+        -- in the floor list, not here.
+        WHERE released_at IS NOT NULL
+          AND ($1::TEXT IS NULL OR client_lineage = $1::TEXT)
+          AND ($2::NUMERIC IS NULL OR available_epoch >= $2::NUMERIC)
```
ochaloup: 🤖 claude: nit — filtering on `available_epoch` drops a release whose epoch could not be resolved even when its `released_at` falls inside the requested window, so `since_epoch` silently hides exactly the rows the calendar failed to place. Worth either falling back to `released_at` or documenting the drop.

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+
+    let mut releases = Vec::new();
+    for fetcher in fetchers {
+        releases.extend(fetcher.fetch()?);
+    }
+
+    // Counted by the row's own source, not by the fetcher: the seed file carries rows from every
+    // source that cannot be fetched.
```
ochaloup: 🤖 claude: nit — there is no releases seed file anywhere in the repo, so this explains the tally with a mechanism that does not exist.

```suggestion
```

api/src/handlers/releases.rs
```diff
@@ -0,0 +1,88 @@
...
+        Err(err) => {
+            error!("Failed to fetch releases: {err}");
+            return Ok(response_error(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                "Failed to fetch records!".into(),
+            ));
+        }
+    };
+
+    let (mut releases, mut sfdp_floors) = (releases, sfdp_floors);
```
ochaloup: 🤖 claude: nit — this rebinds what the `match` just produced; binding `let (mut releases, mut sfdp_floors) = match tokio::try_join!(…)` directly drops the line.

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+pub fn firedancer_lineage(version: &str) -> &'static str {
+    if version.starts_with("0.") {
+        "frankendancer"
+    } else {
+        "firedancer"
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+pub enum FetcherName {
```
ochaloup: 🤖 claude: nit — `FetcherName` mirrors `ReleaseSource` variant-for-variant and forces a hand-written mapping in `collect_releases_info` between two enums that mean the same thing. Deriving `clap::ValueEnum` on `ReleaseSource` lets this one go away.

csv/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+        );
+    }
+
+    #[test]
+    fn a_short_row_is_refused() {
+        assert!(parse_rows("name,epoch\nagave\n").is_err());
+    }
+
+    #[test]
+    fn the_env_var_replaces_the_vendored_table() {
```
ochaloup, line 216: 🤖 claude: nit — this writes a fixed `temp_dir()/ds-csv-override.csv` and mutates process env, so two concurrent `cargo test -p ds-csv` runs collide on the one path (one removes it while the other reads) and the write follows any symlink already sitting there. A pid-unique filename settles both.

csv/src/lib.rs
```diff
@@ -0,0 +1,242 @@
+//! One CSV reader for the tables this workspace carries: a header check, blank-cell validation and
+//! serde row deserialization, so no caller writes its own.
```
ochaloup, line 2: 🤖 claude: nit — "so no caller writes its own" does not hold yet: `api/src/handlers/admin_score_upload.rs` still parses an uploaded CSV with raw `csv::Reader::from_reader`, keeping the headerless-file bug this module documents itself as preventing. Either route that site through `ds-csv` or narrow the claim to the vendored tables.

collect/src/releases/github.rs
```diff
@@ -0,0 +1,113 @@
...
+        if let Some(token) = &self.token {
+            request = request.bearer_auth(token);
+        }
+        Ok(request.send()?.error_for_status()?.json()?)
+    }
+
+    fn fetch_repo(&self, repo: &str, lineage: Option<&str>) -> anyhow::Result<Vec<ReleaseEntry>> {
+        let mut entries = Vec::new();
+
+        for page in 1..=MAX_PAGES {
```
ochaloup, line 65: 🤖 claude: nit — the loop exits silently at `MAX_PAGES`, so a repo past 3000 releases truncates with no signal, while the const's own comment ("a higher count means the loop lost its exit") assumes someone would notice. A `warn!` on cap exit makes it visible.

collect/src/solana_service.rs
```diff
@@ -89,23 +90,30 @@ struct ClientRegistry {
...
 
+const CLIENT_ID_COLUMNS: [Column; 2] = [required("client_id"), required("client_name")];
+
+#[derive(Deserialize)]
+struct ClientIdRow {
+    client_id: u16,
+    client_name: String,
+}
+
+fn parse_client_registry(csv: &str) -> anyhow::Result<ClientRegistry> {
```
ochaloup: 🤖 claude: nit — the parameter shadows the `csv` crate name, so `csv::parse::<ClientIdRow>(csv, …)` on the next line reads as if the crate itself were being indexed. Renaming it to `text` (and the call's first argument with it) removes the double take.

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,230 @@
+use super::{firedancer_lineage, ReleaseEntry, ReleaseFetcher, ReleaseSource};
+use crate::common::retry_blocking;
+use log::{debug, info, warn};
+use serde::Deserialize;
+use std::collections::BTreeMap;
+use std::thread;
+use std::time::Duration;
+
+const SFDP_URL: &str = "https://api.solana.org/api/community/v1/sfdp_required_versions";
```
ochaloup: Shouldn't this be an ARG?

collect/src/releases/github.rs
```diff
@@ -0,0 +1,113 @@
+use super::{firedancer_lineage, normalize_version, ReleaseEntry, ReleaseFetcher, ReleaseSource};
+use crate::common::{retry_blocking, QuadraticBackoffStrategy};
+use crate::solana_service::is_plausible_node_version;
+use chrono::{DateTime, Utc};
+use log::{debug, info, warn};
+use serde::Deserialize;
+
+const GITHUB_API: &str = "https://api.github.com";
```
ochaloup: Shouldn't this be an ARG? (not such a strong opinion on this as for sfdp url)

store/release_feature_gates.csv
```diff
@@ -0,0 +1,19 @@
+# Feature-gate floors: the version the cluster required, and the epoch it started requiring it.
+# Announced in the Solana Tech Discord, served by no API. Rebuilt from
+# marinade-finance/marinade-experiments/explore-solana-versions/discord/, which keeps the messages.
```
ochaloup: Do we have some process for who is responsible for changing this? Shouldn't it rather be part of the ops-infra repo, where we often store configuration that differs between mainnet and testnet?

csv/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+# The package id differs from the lib so `cargo -p` is not ambiguous with the external csv crate;
+# callers still say `csv::parse`.
+name = "ds-csv"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+name = "csv"
```
ochaloup, line 10: Nice of the reuse at the call sites 🤟

ochaloup (changes requested):
I like it. Adding few comments and Claude adding few. Let's discuss the questions in the particular threads.

collect/src/releases/github.rs
ilmari-h: in my opinion this is stable enough to hardcode

store/release_feature_gates.csv
ilmari-h: yes this is here as first example but it can be overridden by ops-infra config. It will be kept up to date in ops-infra instead of here. I will document this in internal docs also.

collect/src/releases/sfdp.rs
ilmari-h: probably this yea I will change it

ilmari-h: while at it might as well change the other one too why not

collect/src/releases/sfdp.rs
ilmari-h: fixed

collect/src/releases/sfdp.rs
ilmari-h: should be stable format but doesn't hurt, added

incident-free-validators.json
ilmari-h: yes test artifact, removing

collect/src/releases/mod.rs
ilmari-h: referring to the csv but this comment probably not needed either way

ilmari-h:
thanks @ochaloup fixed most of the points except for 1 nit I think, ready for another check

collect/src/releases/github.rs
ilmari-h: added arg opt anyway

store/src/releases.rs
```diff
@@ -0,0 +1,460 @@
...
+        FROM {RELEASES_TABLE}
+        -- A row with no timestamp exists only because SFDP named the version as a floor; it belongs
+        -- in the floor list, not here.
+        WHERE released_at IS NOT NULL
+          AND ($1::TEXT IS NULL OR client_lineage = $1::TEXT)
+          AND ($2::NUMERIC IS NULL
+               OR available_epoch >= $2::NUMERIC
+               -- Published after the epoch started, but too recent for the epochs table to place.
+               OR (available_epoch IS NULL
+                   AND released_at >= (SELECT start_at FROM epochs WHERE epoch = $2::NUMERIC)))
```
ochaloup: 🤖 claude: This is the `released_at` fallback added for the last round's `since_epoch` comment, and it misses the case it was added for: when `$2` names an epoch with no `epochs` row — the running epoch, which is exactly when a release has no `available_epoch` — the scalar subquery is NULL, `released_at >= NULL` is NULL, and every unplaced release is dropped. `since_epoch_keeps_a_release_too_recent_to_place` passes 1001, which is in `epochs`, so it does not cover this.

```suggestion
               OR (available_epoch IS NULL
                   AND released_at >= COALESCE(
                       (SELECT start_at FROM epochs WHERE epoch = $2::NUMERIC),
                       (SELECT MAX(end_at) FROM epochs))))
```

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,286 @@
...
+    effective
+}
+
+impl ReleaseFetcher for SfdpFetcher {
+    fn fetch(&self) -> anyhow::Result<Vec<ReleaseEntry>> {
+        let mut agave: BTreeMap<u64, String> = BTreeMap::new();
+        let mut firedancer: BTreeMap<u64, String> = BTreeMap::new();
+        // One epoch of context: a floor already in force here is one whose own start is outside the
+        // window, and must not be reported as starting at the window's edge.
+        let anchor_epoch = self.from_epoch.saturating_sub(1);
```
ochaloup, line 125: 🤖 claude: The anchor closes the ratcheting-floor case from the last round on the common path, but only when `from_epoch - 1` is actually answered — an epoch the endpoint 404s is simply absent from the series, as `a_gap_in_the_series_does_not_start_a_new_run` documents, so the guard below never fires and the inherited floor is again recorded as starting at the window edge. The cron moves the anchor every run, so it lands on a gap eventually; my own suggested fix last round (seed `previous` from the anchor) would not have caught this either.

Walking the anchor back a few epochs until one answers, and keying the guard on the anchor's version rather than its epoch, covers it.

collect/src/releases/sfdp.rs
```diff
@@ -0,0 +1,286 @@
...
+        previous = Some(version);
+        if *epoch == anchor_epoch {
+            continue;
+        }
+        if let Some((_, effective_epoch)) = effective.iter_mut().find(|(v, _)| v == version) {
+            warn!(
+                "Floor {version} is effective again at epoch {epoch}, replacing epoch {effective_epoch}"
+            );
+            *effective_epoch = *epoch;
+            continue;
```
ochaloup, line 111: 🤖 claude: The anchor `continue` above skips the inherited version before it ever reaches `effective`, so when that same version returns later in the window this branch finds nothing to replace and pushes it fresh at the reinstatement epoch — overwriting the backfilled true start. Anchor 1029-1040 answering 4.0.2 (stored start 992), 1041-1045 answering 4.1.0, 1046 back to 4.0.2, and 4.0.2's floor epoch becomes 1046.

csv/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+pub fn load_vendored<T: DeserializeOwned>(
+    default: &'static str,
+    env_var: &str,
+    columns: &[Column],
+    label: &str,
+) -> Result<Vec<T>> {
+    match std::env::var_os(env_var) {
+        Some(path) => load_path(Path::new(&path), columns),
+        None => parse(default, columns, label),
+    }
```
ochaloup, line 96: 🤖 claude: The last round's nit here asked for a pid-unique filename; deleting `the_env_var_replaces_the_vendored_table` instead left this branch — the one ops-infra will use to override `release_feature_gates.csv` — with no test at all. A regression that made `load_vendored` ignore the env var would pass the whole suite and surface as the API serving stale feature-gate floors.

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+                );
+                let current_epoch = client.get_epoch_info()?.epoch;
+                let from_epoch = params
+                    .from_epoch
+                    .unwrap_or_else(|| current_epoch.saturating_sub(params.epochs_back));
+                fetchers.push(Box::new(sfdp::SfdpFetcher::new(
+                    params.sfdp_api_url.clone(),
+                    from_epoch,
+                    // A floor is revisable until its epoch starts.
+                    current_epoch,
```
ochaloup, line 173: 🤖 claude: Dropping `UPCOMING_EPOCHS` closes the last round's retracted-upcoming-floor comment by removing the capability: `/releases` now reports what the floor is and was, never what it is about to become — which is the window in which a validator can still upgrade before falling below it. Worth confirming that is intended rather than a side effect; keeping the lookahead instead needs the retraction clean-up — carry the fetched window in the snapshot and clear `sfdp_floor_epoch` for rows inside it the snapshot no longer names.

migrations/0027-releases.sql
```diff
@@ -0,0 +1,30 @@
...
+  -- As the client reports it in gossip, e.g. 4.2.2, 4.3.0-rc.0, 0.1106.40201, 26.8.2
+  client_version TEXT NOT NULL,
+  -- Epoch the release was published in
+  available_epoch NUMERIC NULL,
+  released_at TIMESTAMP WITH TIME ZONE NULL,
+  release_url TEXT NULL,
+  -- First epoch the Solana Foundation Delegation Program required this version
+  sfdp_floor_epoch NUMERIC NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
```
ochaloup, line 19: 🤖 claude: GEN-8811 asks for `source` and `source_ref` on this table and for precedence that is "deterministic and documented"; neither column is here, so a manually-entered floor cannot be told from a collected one and the next collector run silently overwrites it. Substituting SFDP for the ticket's Discord and anza_wiki sources is an improvement — but either add the column now while the table is new, or update the ticket to record the narrowed model.

collect/README.md
```diff
@@ -56,3 +56,39 @@ cargo run --bin collect -- -u $RPC_URL validators-block-rewards --epoch $EPOCH |
...
+# Steady state: the recent floor window plus the current release lists.
+cargo run --bin collect -- releases > releases.yaml
+
+# One-off historical backfill of every floor the endpoint answers for.
+cargo run --bin collect -- releases --from-epoch 688 > releases.yaml
+```
+
+Adding a source means implementing `ReleaseFetcher` and adding a `--source` value; the table does
+not change. Rows from different sources coexist and are resolved by the precedence documented in
+`migrations/0027-releases.sql`.
```
ochaloup: 🤖 claude: `migrations/0027-releases.sql` documents no precedence — it documents the opposite mechanism, "two collectors fill it and they write disjoint columns", so a reader chasing this pointer finds nothing and a future third source would be written against a rule that was never there.

```suggestion
Adding a source means implementing `ReleaseFetcher` and adding a `--source` value; the table does
not change. The two collectors write disjoint columns — GitHub fills availability, SFDP fills the
floor — so there is no precedence between sources to resolve.
```

api.md
```diff
@@ -183,6 +183,48 @@ curl -sfLS localhost:8000/validators/XkCriyrNwS3G4rzAXtG5B1nnvb5Ka1JtCku93VqeKAr
...
+      "released_at": "2026-08-28T18:47:41Z",
+      "release_url": "https://github.com/anza-xyz/agave/releases/tag/v4.2.2",
+      "updated_at": "2026-09-10T07:11:50.467767Z"
+    }
+  ],
+  "sfdp_floors": [
+    {
+      "client_lineage": "agave",
+      "client_version": "4.2.2",
+      "effective_epoch": 1033
```
ochaloup: 🤖 claude: nit — this floor sits 7 epochs past the same response's `available_epoch` of 1026, which was only reachable while the collector fetched `current_epoch + UPCOMING_EPOCHS`. With `to_epoch = current_epoch` no floor beyond the running epoch is collected any more, so this example can no longer occur; the right value follows whichever way the lookahead question is decided.

ochaloup (dismissed):
I like the changes and thanks for elaboration in the discussion.

I'm putting down a few more findings from Claude coming from the PR update based on the prior review.

The nits can be skipped, but maybe if you can check the changes based on the prior review: `floor ratchets forward`, which is now "replaced" by two other findings — `anchor 404 reopens it` + `reinstated floor overwrites backfilled start`.

collect/src/releases/mod.rs
ilmari-h: we are adding this for historical incidents so future release floors not relevant for the api's consumer

ilmari-h:
@ochaloup thanks, checking these and also figured out a way to programmatically keep the feature gate floors up to date by reading a couple JSON endpoints published by Anza. Adding that as well soon so gonna need another review

store/src/releases.rs
ilmari-h: fixed

collect/src/releases/sfdp.rs
ilmari-h: 🤖 Not fixing. Debug scan of 686..=1032: no floor for `681-685, 687, 688` only. The anchor lands in that hole solely at `from_epoch = 689`; the real backfill starts at 686, where the walk finds nothing and today's code is already right, and cron sits ~20 epochs off the tip. Cost if it ever fired: one wrong epoch in the 680s, outside the ~80-epoch API window.

collect/src/releases/sfdp.rs
ilmari-h: 🤖 Valid mechanically: the anchor seeds `previous` without pushing to `effective`, so a reinstatement of that version pushes fresh at the later epoch. Needs SFDP to lower the floor back to a pre-window version — never happened in 350 epochs, the `warn!` has never fired. Same family as the open-by-design entry in `comments.md` (one row per version cannot hold two runs). Not fixing; the 3-line guard is to remember the anchor's version and skip it if it reappears, if this ever shows up.

csv/src/lib.rs
ilmari-h: 🤖 Fair — the branch is live via `OPERATORS_CSV`, which ops-infra sets in the API deployment. Test re-added with a per-test env var name and a pid-unique path, which is what the original nit wanted, plus one for the unset case. Both branches of `load_vendored` covered now.

migrations/0027-releases.sql
ilmari-h: 🤖 Divergence is deliberate: the table is keyed on `(client_lineage, client_version)`, one row per version, and the fetchers write disjoint columns — GitHub availability, SFDP and the tracker one floor each — so there is no precedence left to document. Adding `source` back means one row per source, which is what the key replaced. The real gap you name is that a seeded floor is indistinguishable from a derived one and the collector overwrites it; that is the chosen model (tracker and chain are the source of truth) and it is now recorded in the PR description rather than the schema. Ticket text is narrower than reality — noted, not changing the table for it.

collect/README.md
ilmari-h: 🤖 Right, the pointer was stale. Reworded: the fetchers write disjoint columns so no source overrides another, and the one ordering that does exist is within the feature-gate column — a derived floor overwrites the epoch the migration seeded. Kept it at three fetchers rather than two.

ilmari-h: dropped this commentary, it was not needed

collect/src/releases/feature_gates.rs
```diff
@@ -0,0 +1,384 @@
...
+        .filter_map(|value| value.parse::<ValidatorVersion>().ok())
+        .min()
+}
+
+/// Whether a gate could raise the floor: only one requiring at least the published floor can, and
+/// the published floor is the maximum over every gate already activated.
+fn could_raise_floor(gate: &Gate, published: &BTreeMap<&str, ValidatorVersion>) -> bool {
+    LINEAGES.iter().any(
+        |lineage| match (gate.version_for(lineage), published.get(*lineage)) {
+            (Some(required), Some(floor)) => required >= *floor,
```
ochaloup, line 67: 🤖 claude: Bounding the read set by the *live* published floor means a step is unrecoverable once Anza moves past it — miss the window in which `current` is 4.2.0 and no future run can ever reach that gate, so `get_feature_gate_floor_at_epoch(agave, 1030)` answers 4.1.0-beta.1 from epoch 1008 and only a migration edit repairs it. `check_published_floor_was_derived` cannot catch this: it asserts the current floor is present, which the filter guarantees by construction.

Reading every gate instead costs three `getMultipleAccounts` calls rather than one (252 gates, `ACCOUNTS_PER_CALL = 100`) and is strictly more correct by the argument the removed doc comment made — no subset can omit a higher gate that activated earlier. Failing that, `--latest-gates-to-read` was the escape hatch and it went away in 3de89be.

store/src/releases.rs
```diff
@@ -0,0 +1,479 @@
...
+        let effective_epoch: u64 = row.get::<_, Decimal>("effective_epoch").try_into()?;
+        let Ok(version) = version.parse::<ValidatorVersion>() else {
+            warn!("Floor row {lineage} {version} is not a version, skipping it");
+            continue;
+        };
+
+        newest
+            .entry(lineage)
+            .and_modify(|floor| {
+                if (effective_epoch, &version) > (floor.0, &floor.1) {
```
ochaloup, line 468: 🤖 claude: Generalising this over both floor columns dropped the `updated_at DESC` tiebreak the previous ordering had, and version-wins is the wrong rule for the only case that produces an SFDP tie. `upsert_floors` conflicts on `(client_lineage, client_version)` and never deletes, so when SFDP revises the still-open epoch 1033 down from 4.2.2 to 4.2.1 both rows survive at that epoch and this comparison answers 4.2.2 — the version that was retracted.

Selecting `updated_at` and ordering `(effective_epoch, updated_at, version)` is correct for both columns: a correction or a re-derivation always carries the newer timestamp, and a migration-seeded row carries the oldest.

store/src/releases.rs
```diff
@@ -0,0 +1,479 @@
...
+        -- A row with no timestamp exists only because a floor named the version; it belongs in a
+        -- floor list, not here.
+        WHERE released_at IS NOT NULL
+          AND ($1::TEXT IS NULL OR client_lineage = $1::TEXT)
+          AND ($2::NUMERIC IS NULL OR released_at >= (
+              -- The epoch asked for may have no row: older than the history we keep takes
+              -- everything, newer than it -- the running epoch -- takes what followed the last
+              -- close.
+              SELECT CASE
+                  WHEN $2::NUMERIC <= (SELECT MIN(epoch) FROM epochs) THEN '-infinity'::TIMESTAMPTZ
```
ochaloup, line 277: 🤖 claude: nit — the comment's own justification is "the epoch asked for may have no row", but an epoch equal to `MIN(epoch)` has one and the `ELSE` branch answers it correctly; `<=` hands it to `'-infinity'` instead. Checked on postgres 16 with epochs 1000-1002 and a release published seven months before epoch 1000: `since_epoch=1000` returns it, `<` does not, and every other case is byte-identical.

```suggestion
                  WHEN $2::NUMERIC < (SELECT MIN(epoch) FROM epochs) THEN '-infinity'::TIMESTAMPTZ
```

store/src/releases.rs
```diff
@@ -0,0 +1,479 @@
...
+                  ELSE COALESCE(
+                      (SELECT MIN(start_at) FROM epochs WHERE epoch >= $2::NUMERIC),
+                      (SELECT MAX(end_at) FROM epochs)
+                  )
```
ochaloup, line 281: 🤖 claude: nit — this fallback exists for the running epoch, but it fires for every epoch past the last close, so with the table ending at 1002 a `since_epoch=2000` request is answered with the same bound as `since_epoch=1003` and returns everything published since 1002. `since_epoch` is an unvalidated `Option<u64>` on the public query params, so any caller can reach it.

Bounding it to `MAX(epoch) + 1` and returning nothing beyond that would keep all three boundaries decided in this one `CASE`.

store/src/releases.rs
```diff
@@ -0,0 +1,479 @@
...
+                ],
+            )
+            .await?;
+    }
+
+    Ok(total)
+}
+
+/// What was published: one row per version the client's releases carry a timestamp for.
+/// Writes only the feature-gate floor.
```
ochaloup, line 199: 🤖 claude: nit — the first line is a mis-paste from the availability upsert: it describes rows carrying `released_at`, which is exactly what this function does not write. Its two siblings each carry one accurate line.

```suggestion
/// Writes only the feature-gate floor.
```

collect/src/releases/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+                )?));
+            }
+        }
+    }
+
+    let mut releases = Vec::new();
+    for fetcher in fetchers {
+        releases.extend(fetcher.fetch()?);
```
ochaloup, line 181: 🤖 claude: nit — aborting the whole snapshot is the right default, but the order makes it expensive: the default `--source` is `github,sfdp,feature-gates`, so the paced SFDP pass (one request per epoch, ~4 minutes for a `--from-epoch 688` backfill) runs first and is thrown away when the wiki fetch fails. That fetch is the likeliest to fail — `Gate` and `VersionFloorFile` deserialize renamed keys with no `#[serde(default)]` from a page repo collaborators edit.

Validating both JSONs in `FeatureGatesFetcher::new` would fail in the first second instead, with no error-swallowing anywhere.

collect/src/releases/feature_gates.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    /// The epoch each gate activated in, absent for one still pending.
+    fn activation_epochs(&self, gates: &[&Gate]) -> anyhow::Result<BTreeMap<String, Epoch>> {
+        let epoch_schedule = self.rpc.get_epoch_schedule()?;
+        let mut epochs = BTreeMap::new();
+
+        for chunk in gates.chunks(ACCOUNTS_PER_CALL) {
+            let ids: Vec<Pubkey> = chunk
+                .iter()
+                .filter_map(|gate| Pubkey::from_str(&gate.feature_id).ok())
```
ochaloup, line 141: 🤖 claude: nit — a `Feature ID` that does not parse is dropped without a word, so the floor step it represents vanishes from the timeline with nothing to trace it to. Every other rejection in this module says something (`debug!("Gate {id} holds no feature yet")` two lines down), and the input is a community-editable wiki page, which is where a typo actually lands.

collect/src/releases/feature_gates.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// silent everywhere else.
+    fn check_published_floor_was_derived(
+        &self,
+        published: &BTreeMap<&str, ValidatorVersion>,
+        derived: &[ReleaseEntry],
+    ) {
+        for (lineage, floor) in published {
+            if !derived
+                .iter()
+                .any(|entry| entry.client_lineage == *lineage && entry.client_version == *floor)
```
ochaloup, line 244: 🤖 claude: nit — exact membership fails on correct output whenever a gate requiring a higher version activated earlier, because `floor_timeline` then never emits the published version as its own step. `a_gate_needing_an_older_version_does_not_lower_the_floor` documents that gates activate out of order, and `version-floor.json` carries a `next` field precisely because `current` lags the chain.

Asserting the highest derived version is at or above the published floor is the relation that actually holds, and keeps the module's only signal trustworthy.

collect/src/solana_service.rs
```diff
@@ -209,21 +217,8 @@ impl ClientId {
...
-            Some((patch, prerelease)) => {
-                numeric(patch)
-                    && !prerelease.is_empty()
-                    && prerelease
-                        .bytes()
-                        .all(|b| b.is_ascii_alphanumeric() || b == b'.')
-            }
-        })
+pub fn is_plausible_node_version(version: &str) -> bool {
+    crate::validator_version::ValidatorVersion::from_gossip(version).is_ok()
```
ochaloup, line 221: 🤖 claude: nit — this parses and throws the parse away, so the caller stores the raw gossip string while `releases` stores `ValidatorVersion::as_str()`. That leaves the type's own invariant ("one version has one spelling wherever it is stored") asserted but not enforced, and the join those two tables exist for — the late-client-version incident — would miss silently on any drift.

Returning `Option<ValidatorVersion>` and storing `version.as_str()` at the call site is one line and makes the invariant true. (I checked the rewrite against the predicate it replaced and found no behaviour change that matters.)

csv/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    #[test]
+    fn the_env_var_replaces_the_vendored_table() {
+        // What ops-infra does with OPERATORS_CSV: mounts a file and points the binary at it. The
+        // var name and the path are unique to this test, so it shares no state with another.
+        let var = "DS_CSV_TEST_OVERRIDE_REPLACES_VENDORED";
+        let path = std::env::temp_dir().join(format!("ds-csv-{}-{var}.csv", std::process::id()));
+        std::fs::write(&path, "epoch,name\n1019,firedancer\n").unwrap();
+        std::env::set_var(var, &path);
```
ochaloup, line 222: 🤖 claude: nit — the per-test variable name and pid-unique path are exactly right, but `set_var`/`remove_var` are not thread-safe against a concurrent `getenv`, and `the_vendored_table_is_used_when_the_env_var_is_unset` calls `load_vendored` (which is `env::var_os`) on another harness thread. The two `unwrap()`s after `set_var` also leak the variable and the temp file into the rest of the run on a failure.

Splitting the env lookup out of `load_vendored` and testing the `Option<&OsStr>` half removes both the filesystem and the process env from the test.

ochaloup (approved):
The Claude came with findings coming from the changes done based on the prior review round.
Here I think they are rather minor or nitpicks so approving, if you may double check, feel free to decline them.

collect/src/releases/feature_gates.rs
ilmari-h: this is unlikely to happen as anza doesnt typically publish multiple of these floors a day or anything like that, dismissing

PR #275, opened 2026-09-10
[GEN-8946][commission] fixing simd commission changes from solana

ochaloup:
Restores `commission_effective`, which SIMD-0232 left null for the entire fleet from epoch 1030, by sampling it from raw vote state — and collects the vote-state v4 commission and collector fields the API had no way to see before.

- Parse vote accounts directly, handling v1_14_11, V3 and v4 side by side, since `getVoteAccounts` has never exposed a collector address and reward rows stopped carrying commission at all
- Fall back to the sampled vote state when a reward row has no commission, and record `commission_effective_source` beside it so a consumer can tell an applied rate from a sampled one
- Persist and expose the collector addresses, basis points, redirect flags and shared-collector counts per epoch, so it is visible where a validator's commission is actually paid without diffing pubkeys
- Repair the high-commission warning and the rug detector, both of which fell silent the moment the applied rate went null
- Stop APR/APY reading a missing commission as 100%, which had zeroed yields fleet-wide for epochs 1030 and 1031
- Track validators by commission source in Prometheus, so the next upstream change of this kind surfaces before a consumer reports it

## Summary by CodeRabbit

- **New Features**
  - Added validator inflation-reward and block-revenue commission details.
  - Added collector addresses, commission versions, shared-collector information, and pending delegator rewards to validator records.
  - Commission data now identifies whether values came from reward data or vote-state information.
  - Added metrics showing validator counts by inflation commission source, including unknown sources.

- **Bug Fixes**
  - Improved handling of older and newer vote-account formats.
  - Preserved existing validator details when updated data is incomplete.
  - Improved high-commission and rug-detection accuracy when commission data is unavailable.
  - Ensured metrics refresh consistently without publishing data for unseen records.


store/src/utils.rs
```diff
@@ -405,33 +400,33 @@ pub async fn load_ruggers(psql_client: &Client) -> anyhow::Result<HashMap<String
...
                     OR
-                    (prev_commission <= 10 AND commission_effective > 10 AND next_commission <= 10)
+                    (prev_commission <= 10 AND commission_max_observed > 10 AND next_commission <= 10)
             )
             SELECT
                 vote_account,
                 COUNT(*) AS events_count,
                 ARRAY_AGG(epoch) AS epochs,
-                ARRAY_AGG(commission_effective) AS commission_observed_values,
+                ARRAY_AGG(commission_max_observed) AS commission_observed_values,
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — `ORDER BY epoch` added to all three `ARRAY_AGG` calls. The index zip was already safe, but the exposed order was not guaranteed, and the rugger test's ascending-epoch assertion was resting on the query plan rather than on SQL.

store/src/utils.rs
```diff
@@ -405,34 +400,34 @@ pub async fn load_ruggers(psql_client: &Client) -> anyhow::Result<HashMap<String
...
                     commission_changes
                 WHERE
-                    (commission_effective > commission_min_observed AND commission_effective > 10 AND commission_min_observed <= 10)
+                    (commission_advertised > commission_min_observed AND commission_advertised > 10 AND commission_min_observed <= 10)
                     OR
-                    (prev_commission > 10 AND commission_effective <= 10 AND next_commission > 10)
+                    (prev_commission > 10 AND commission_advertised <= 10 AND next_commission > 10)
                     OR
-                    (prev_commission <= 10 AND commission_effective > 10 AND next_commission <= 10)
+                    (prev_commission <= 10 AND commission_advertised > 10 AND next_commission <= 10)
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — `filtered_commissions` now gates on `commission_min_observed IS NOT NULL`, since branches 2 and 3 never read the floor and `close_epoch` writes it only at close. A regression test reproduces the `Vec<i32>` decode panic without that gate.

collect/src/solana_service.rs
```diff
@@ -508,30 +508,162 @@ fn extract_json_value(json: &Map<String, Value>, key: String) -> Option<String>
...
+            )?)),
+            pending_delegator_rewards: Some(u64::from_le_bytes(read_array::<8>(
+                data,
+                VOTE_V4_PENDING_DELEGATOR_REWARDS_OFFSET,
+                "pending_delegator_rewards",
+            )?)),
+        }),
+        _ => anyhow::bail!("unsupported vote state version {version}"),
+    }
+}
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — `bps_to_percent` is `bps.min(10_000).div_ceil(100)` and is the only path from bps to a percent, so `commission_effective` cannot exceed 100, and the warnings read whole-percent columns rather than bps. The stored column is deliberately the raw vote-state value, matching agave and solana-snapshot-parser.

migrations/0028-vote-state-v4-commission-collectors.sql
```diff
@@ -0,0 +1,33 @@
...
+  ADD COLUMN block_revenue_commission_bps INTEGER DEFAULT NULL,
+  ADD COLUMN pending_delegator_rewards NUMERIC DEFAULT NULL,
+  -- 'reward_row' where the runtime told us the rate it applied, 'vote_state' where SIMD-0232 left
+  -- only the sampled state to go on. NULL for epochs closed before this column existed and for a
+  -- validator with neither source, which is not the same as a rate of zero.
+  ADD COLUMN commission_effective_source TEXT DEFAULT NULL;
+
+-- The collector is many-to-one by design and needs no signature of its own, so lookups group by it.
+CREATE INDEX validators_inflation_rewards_collector ON validators (epoch, inflation_rewards_collector);
+CREATE INDEX validators_block_revenue_collector ON validators (epoch, block_revenue_collector);
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — both indexes were dropped from the migration in an earlier round because no query in the repo could use them, so no index build is left to block writes.

migrations/0028-vote-state-v4-commission-collectors.sql
```diff
@@ -0,0 +1,29 @@
+-- Vote state v4 (SIMD-0185) commission and collector fields, sampled from raw vote accounts.
+-- Names and null semantics match solana-snapshot-parser's ValidatorMeta so this derivation and the
+-- one stakes-etl loads into BigQuery stay reconcilable.
+--
+-- Basis points are the authoritative rate; commission_advertised / commission_effective /
+-- commission_max_observed / commission_min_observed stay whole-percent projections of it.
+--
+-- Nothing here backfills commission_effective, and from epoch 1030 on nothing can: the epoch-1030
```
ilmari-h: nit: perhaps over-commented by claude?

collect/src/solana_service.rs
```diff
@@ -508,30 +508,183 @@ fn extract_json_value(json: &Map<String, Value>, key: String) -> Option<String>
...
+        warn!(
+            "{without_commission} vote accounts hold a version this build reads no commission from"
+        );
+    }
+    info!(
+        "Parsed {} vote accounts, {v4} on vote state v4",
+        states.len()
+    );
+
+    Ok(states)
```
ilmari-h, line 691: ### 🤖 Mass parse failure degrades silently to zero self-stake

`get_vote_account_states` returns `Ok` no matter how many accounts failed to parse — the only signal is the `warn!` above. The code it replaces returned `Err` on a bad read.

So a cluster-wide vote-state shape change lands as: `states` empty → `withdraw_authorities` empty → `self_stake` 0 for the whole fleet, which feeds eligibility and scoring. Exactly the failure mode the new `ds_validator_inflation_commission_source` gauge exists to catch downstream, except this one happens a stage earlier and nothing exports it.

Suggest bailing when `unparsed` exceeds a fraction of `vote_accounts.len()` — a handful of unparsed accounts is normal, all of them is not.

store/src/utils.rs
```diff
@@ -391,12 +391,7 @@ pub async fn load_versions(
 
     Ok(records)
 }
-/*
-We are checking if:
-- Current commission is greater than previous minimum, and it's above 10 OR
-- Previous commission is more than 10, current commission is less than or equal to 10, and the next commission is more than 10 OR
-- Previous commission is less than or equal to 10, current commission is more than 10, and the next commission is less than or equal to 10 OR
- */
+// Reads the epoch's last advertised rate, not commission_effective: SIMD-0232 nulled that, and SQL comparisons against NULL are never true, so no epoch-1031-or-later row could match. It stands in because the runtime charges the rate held at the epoch's last slot, which is what commission_effective recorded; commission_max_observed would not, as a ceiling cannot tell a rug from an honest cut - both leave the same floor and ceiling.
```
ilmari-h: ### 🤖 Rugger switch to commission_advertised re-scores all history

The query has no epoch cutoff, so moving `commission_effective` → `commission_advertised` re-evaluates every epoch ever recorded, not just the ones SIMD-0232 nulled. Validators currently flagged `rugged_commission` can un-flag, and validators never flagged can appear — from a deploy that changed no data.

If that restatement is intended, it belongs in the PR body. If not, keep `commission_effective` where it exists and fall back per row, e.g. `COALESCE(commission_effective, commission_advertised)`, so closed epochs up to 1029 keep the rate the runtime actually applied.

store/src/utils.rs
```diff
@@ -405,34 +400,38 @@ pub async fn load_ruggers(psql_client: &Client) -> anyhow::Result<HashMap<String
...
                 FROM
                     commission_changes
                 WHERE
-                    (commission_effective > commission_min_observed AND commission_effective > 10 AND commission_min_observed <= 10)
-                    OR
-                    (prev_commission > 10 AND commission_effective <= 10 AND next_commission > 10)
-                    OR
-                    (prev_commission <= 10 AND commission_effective > 10 AND next_commission <= 10)
+                    -- Branches 2 and 3 never read the floor, which close_epoch writes only at close
+                    commission_min_observed IS NOT NULL
```
ilmari-h, line 420: ### 🤖 Floor guard hoisted over branches that never read the floor

The comment says branches 2 and 3 never read the floor — which is the reason this guard should sit inside branch 1, not outside the `OR`. As written it gates all three, so branches 2 and 3 additionally require `commission_min_observed` to be non-NULL, i.e. a closed epoch row.

```sql
(commission_advertised > commission_min_observed AND commission_min_observed IS NOT NULL AND commission_advertised > 10 AND commission_min_observed <= 10)
OR
(prev_commission > 10 AND commission_advertised <= 10 AND next_commission > 10)
OR
(prev_commission <= 10 AND commission_advertised > 10 AND next_commission <= 10)
```

(The `> ... AND ... <= 10` comparisons are already NULL-safe on their own, so branch 1 arguably needs no explicit guard either.)

store/src/close_epoch.rs
```diff
@@ -16,6 +18,48 @@ pub struct CloseEpochParams {
 
 const DEFAULT_CHUNK_SIZE: usize = 500;
 
+// A reward row still wins where one exists, so pre-1031 epochs reprocess to the same values.
+fn resolve_commission_effective(
```
ilmari-h, line 22: ### 🤖 div_ceil rounding leaks into the observed floor and ceiling

The `vote_state` branch writes a `div_ceil` percent, and `create_epoch_record` then folds that same value into `commission_max_observed = GREATEST(commission_max, commission_advertised, commission_effective)` and `commission_min_observed = LEAST(...)`. A validator at 733 bps records an *observed* ceiling of 8, and that 8 outlives the epoch.

`commission_effective_source` marks the rounded value where it is stored, but the max/min columns carry no such marker and are read by `scoring.rs`, `groups.rs`, `take_rates.rs` and the rugger query. Worth either excluding a `vote_state`-sourced effective rate from the GREATEST/LEAST, or noting in the migration that the observed bounds became approximate from epoch 1030 on.

store/src/utils.rs
```diff
@@ -1046,7 +1066,9 @@ pub async fn load_validators(
             let (apr, apy) = if let Some(c) = apy_calculators.get(&epoch) {
                 let (apr, apy) = c.estimate_yields(
                     row.get::<_, Decimal>("credits").try_into().unwrap(),
+                    // SIMD-0232 left commission_effective null for epochs 1030-1031, and nothing can backfill them; falling straight to 100 would read those epochs as zero yield fleet-wide
```
ilmari-h: ### 🤖 APY fallback restates previously served numbers

Two things here:

1. `.or(...)` is eager — `row.get::<_, Option<i32>>("commission_advertised")` runs for every row regardless of whether `commission_effective` was present. `.or_else(|| row.get(...))` on the hottest loop of the cache warm.
2. This changes APR/APY already served for epochs 1030-1031, and does it silently. `commission_advertised` is the last hourly sample rather than the applied rate, so the new numbers are an estimate where the old ones were an (admittedly null-shaped) fact. Consumers that cached the earlier values will disagree with the API after deploy.

store/src/utils.rs
```diff
@@ -980,6 +982,24 @@ pub async fn load_validators(
...
+                block_revenue_collector,
+                block_revenue_commission_bps,
+                pending_delegator_rewards,
+                -- A vote account's own address never changes and only UpdateCommissionCollector moves this, so a mismatch is a deliberate redirect
+                CASE WHEN inflation_rewards_collector IS NOT NULL
+                     THEN inflation_rewards_collector <> validators.vote_account END AS inflation_rewards_collector_redirected,
+                -- Compared against the identity, its default: agave stopped re-syncing the two once SIMD-0232 activated, so a changed identity also lands here
+                CASE WHEN block_revenue_collector IS NOT NULL
+                     THEN block_revenue_collector = validators.identity END AS block_revenue_collector_is_identity,
+                -- Partitioning on a null collector would pool every pre-v4 validator into one bogus count, so the CASE nulls those rows out instead
```
ilmari-h: ### 🤖 Two window functions added to the cache-warm query

`COUNT(*) OVER (PARTITION BY validators.epoch, inflation_rewards_collector)` and its block-revenue twin partition differently from each other and from the final `ORDER BY epoch DESC`, so Postgres adds two more sorts of the full `display_epochs` × all-validators result set — the single heaviest query in the service.

The counts themselves are correct (the partition includes `epoch`, and the `LEFT JOIN`s are all `DISTINCT ON`-deduped, so no row multiplication). It is purely a cost question: worth an `EXPLAIN ANALYZE` against production-sized data before merge, and if it bites, the same numbers come cheaply from a `GROUP BY epoch, collector` CTE joined back on.

ilmari-h (commented):
Leaving a few AI comments I think can be worth checking

store/src/utils.rs
ochaloup: 🤖 claude: Fixed — the CTE now reads COALESCE(commission_effective, commission_advertised), so advertised stands in only for the epochs SIMD-0232 nulled and no closed epoch is restated. A regression test covers a validator whose spike lands between hourly samples, which the advertised-only series scored as steady.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Fixed — get_vote_account_states now fails the run past 50% unparsed instead of returning a partial set. The premise is narrower than stated though: an unknown version hits the catch-all arm and still yields the withdrawer, and a total wipe-out already panics on the assert in get_self_stake — only the partial case was unguarded.

store/src/utils.rs
ochaloup: 🤖 claude: Declining the move — the guard is hoisted on purpose, so branches 2 and 3 cannot select an open-epoch row whose NULL floor reaches ARRAY_AGG and panics the Vec<i32> decode; load_ruggers_skips_a_matching_epoch_whose_floor_is_not_yet_known pins exactly that. The comment was the real defect and now states the consequence rather than the premise.

store/src/close_epoch.rs
ochaloup: 🤖 claude: Declining — master already routes reward-row commission_bps through the same bps_to_percent, so the rounding has been in those columns since SIMD-0291; restoring commission_effective for 1030-1031 makes them consistent with epochs up to 1029 rather than approximate. Documented on commission_max_observed in dto.rs instead of changing the GREATEST/LEAST, which would reopen the hole this PR closes.

store/src/utils.rs
ochaloup: 🤖 claude: Fixed — .or_else, so the second Row::get no longer runs on every row of the cache warm. Point 2 stands: the description covers the 1030-1031 restatement and is being extended to say the fallback is the epoch last advertised sample rather than the applied rate.

store/src/utils.rs
ochaloup: 🤖 claude: Keeping them — measured at production scale (380k rows, 90 epochs x ~1955 validators) the window form runs 1559ms best against 1149ms with the columns dropped, while the GROUP BY CTE comes in at 1608ms, slower than what it would replace. It is also one extra full sort rather than two: both partitions lead with epoch, so the second arrives via an in-memory Incremental Sort, and 420ms sits on a 600s cache-warm timer.

migrations/0028-vote-state-v4-commission-collectors.sql
ochaloup: 🤖 claude: Fixed — you were right, the migration is down to 5 single-line comments from 22; the SIMD-0232 narrative belongs in the PR description, not in the file. The same pass ran over the whole branch: 26 comments over the 100-char line budget and 11 stacked multi-line blocks, now none of either.

ochaloup:
@ilmari-h May you run a second round of review on this, please?

ilmari-h (dismissed):

ilmari-h (approved):

PR #276, opened 2026-09-14
[GEN-9063] more client and provider data

ilmari-h:
`/clients` and `/providers` endpoints now serve more details:

each row in `/providers` gets
- ASNs
- data center locations breakdown and stake distribution therein
- clients list stake weighted breakdown (Agave x%, Frankendancer y% etc.)
- superminority validators count

each row in `/clients` gets
- latest release version
- stake weighted version spread (v2.3 80%, v2.3-alpha 10% etc.)
- list of block engines
- datacenter country count

## Summary by CodeRabbit

* **New Features**
  * Provider and client group results now include hosting providers, locations, client lineages, versions, stake shares, and client distributions.
  * Client group views now show block-engine information and the latest published client releases.
  * Group listings provide separate provider and client group views with enriched records.
  * Group summaries more accurately represent validator client mixes and version distributions, including validators without recent crawl data.
  * API documentation now includes schemas for releases, locations, sharing details, and enriched validator group records.


api/src/utils/validator_groups.rs
```diff
@@ -434,26 +450,10 @@ mod tests {
         );
     }
 
-    #[test]
-    fn the_unclassified_group_is_searchable_by_name() {
```
ochaloup, line 438: 🤖 claude: nit — this was the only test that searches a client tree by the parent's own name. `matches_query` still checks `node.group.key`, so if that check were dropped, `/clients?query=unknown` would return nothing and every test would still pass.

store/src/groups.rs
```diff
@@ -1186,13 +1719,12 @@ mod tests {
...
         );
         assert_eq!(
-            clients.groups.first().map(|group| group.stake_delta_7d),
+            clients.first().map(|group| group.stake_delta_7d),
             all.clients
                 .nodes
                 .first()
                 .map(|node| node.group.stake_delta_7d)
         );
-        assert_eq!(clients.current_epoch, all.clients.current_epoch);
```
ochaloup, line 1195: 🤖 claude: nit — with this assert and the `providers.current_epoch` one gone, no test checks the `current_epoch` that `/providers` and `/clients` serve, and three places now build it by hand. Asserting `all.providers.current_epoch` and `all.clients.current_epoch` equal `Some(CURRENT_EPOCH)` would restore that coverage.

store/src/groups.rs
```diff
@@ -45,7 +50,7 @@ fn normalized(value: Option<String>) -> Option<String> {
     Some(value.to_string())
 }
 
-fn is_unknown_placeholder(value: &str) -> bool {
+pub fn is_unknown_placeholder(value: &str) -> bool {
```
ochaloup, line 53: 🤖 claude: nit — nothing outside `groups.rs` calls this, so making it `pub` only adds public API.

```suggestion
fn is_unknown_placeholder(value: &str) -> bool {
```

store/src/dto.rs
```diff
@@ -646,6 +646,22 @@ pub struct FeatureSetStats {
...
 
+/// Validator stats on the level of a datacenter location.
+#[derive(Deserialize, Serialize, Debug, Clone, Default, PartialEq, utoipa::ToSchema)]
+pub struct GroupLocation {
+    pub country: String,
+    pub city: Option<String>,
+    pub validator_count: u64,
+    pub total_stake: Decimal,
+}
+/// A published client release, as `/releases` serves it.
```
ochaloup, line 657: 🤖 claude: nit — `/releases` serves `ReleaseRecord` (`client_version`, `release_url`, …), not this shape, and utoipa copies this doc into the OpenAPI schema.

```suggestion
/// The newest release of a client lineage.
```

store/src/groups.rs
```diff
@@ -395,6 +583,7 @@ impl<'a> Population<'a> {
 /// not show up as a loss.
 pub fn aggregate_operators(validators: &[&ValidatorRecord]) -> ValidatorGroups {
     match Population::over(validators.to_vec(), validators.to_vec()) {
+        // Operator rows carry no information panel, so there is nothing to overlay onto them.
```
ochaloup, line 586: 🤖 claude: nit — there is no overlay step anymore. Operator rows get no panel because `aggregate_kind` folds with `Accumulator::<()>`, so this comment sends readers looking for code that doesn't exist.

```suggestion
```

api/src/cache.rs
```diff
@@ -409,6 +410,22 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
     )
     .await?;
 
+    // Load newest releases per lineage.
```
ochaloup, line 413: 🤖 claude: nit — this comment repeats the block below it. The real constraint goes unsaid: `or_insert` keeps the newest release only because `load_releases` sorts by `released_at DESC`.

```suggestion
```

api/src/cache.rs
```diff
@@ -409,6 +410,22 @@ pub async fn warm_validators_cache(context: &WrappedContext) -> anyhow::Result<(
...
+    let releases = {
+        let psql_client = &context.read().await.psql_client;
+        let mut latest = ClientReleases::default();
+        for release in store::releases::load_releases(psql_client, None, None).await? {
+            latest
+                .entry(release.client_lineage.to_lowercase())
+                .or_insert(ClientRelease {
```
ochaloup, line 420: 🤖 claude: This keeps the most recently published tag per lineage, prereleases included. As of today agave shows `4.3.0-rc.1`, firedancer `26.9.2` and frankendancer `0.1202.40302`, all pre-releases, instead of `4.2.2` / `26.8.4` / `0.1106.40201`. Picking the highest stable `ValidatorVersion` also needs the GitHub `prerelease` flag stored: Firedancer marks `v26.09.x` only with that flag, while Agave leaves it `false` on its `-rc` tags.

ochaloup (dismissed):
The code seems ok to me. Claude highlighted few nits so posting those here to consider them. The most relevant seems the cache ordering, otherwise feel free to decline and merge.

store/src/groups.rs
ilmari-h: this is quite trivial and was removed for that reason so wontfix

api/src/utils/validator_groups.rs
ilmari-h: bringing this back

store/src/groups.rs

api/src/cache.rs
ilmari-h: I don't see harm in having the pre-releases there as well. Filtering those out would add complexity to the releases collector and require new database column. We can add this later if needed but leaving out of scope here.

ilmari-h:
@ochaloup thanks, fixed some of the points and responded to rest, asking for another approval

api/src/utils/validator_groups.rs
ochaloup: 🤖 claude: Fixed — `the_unclassified_client_is_a_row_like_any_other` now searches by the parent's own name, so dropping the `node.group.key` check in `matches_query` fails a test. Passing `UNKNOWN_GROUP` rather than a lowercase literal also exercises the lowercasing in `search_term`, which the deleted test did not.

store/src/groups.rs
ochaloup: 🤖 claude: Agreed, closing — `Population::new(...).current_epoch` is already asserted in `epoch_resolution_picks_the_newest_epoch_that_had_ended_by_the_cutoff`, so only the three `current_epoch: Some(...)` plumbing lines are uncovered and a dropped field would not compile.

store/src/groups.rs
ochaloup: 🤖 claude: Fixed — `is_unknown_placeholder` is private again, and `cargo clippy --workspace --all-targets` is clean, so the one in-file caller still reaches it.

store/src/dto.rs
ochaloup: 🤖 claude: Fixed — the doc now reads `The newest release of a client lineage.`, so the OpenAPI schema no longer promises the field names `/releases` serves.

store/src/groups.rs
ochaloup: 🤖 claude: Fixed — the comment is gone, so nothing sends a reader looking for an overlay step that no longer exists.

api/src/cache.rs
ochaloup: 🤖 claude: Fixed — the comment is gone. The constraint it should have carried instead, that `or_insert` is correct only because `load_releases` sorts by `released_at DESC`, is now written nowhere; that belongs on the release-fold thread rather than here.

api/src/cache.rs
ochaloup: 🤖 claude: The prerelease half needs the column, agreed — but the ordering half does not: `or_insert` keeps whichever row `ORDER BY released_at DESC` (`store/src/releases.rs:284`) returned first, so an Agave backport published after a newer line makes `latest_release` walk backwards, and that dependency is now stated nowhere. Folding by max `ValidatorVersion` inside `store::releases` fixes it with no schema change and no collector work — `store` already depends on `collect` and parses `ValidatorVersion` at `store/src/releases.rs:460`.

ochaloup (approved):
All nice. One thread is left open, but rather a minor point.

PR #277, opened 2026-09-15
[GEN-9100] commission spike incident type

ilmari-h:
Added new incident type `CommissionSpike`. Recorded when a validator raised their inflation commission to 90% or more within an epoch (or across epoch boundary).

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting and displaying validator commission spikes of 90% or higher.
  * Commission spike incidents show previous and highest subsequent commission rates, timestamp, and epoch details.
  * Incident filtering now includes commission-spike events alongside downtime and block-production incidents.
  * Commission spikes are evaluated using prior-epoch context while reporting changes within the relevant epoch.
  * MEV and block-revenue commission changes are excluded from commission-spike incidents.


migrations/0030-commissions-epoch-index.sql
```diff
@@ -0,0 +1,2 @@
+-- The only index on commissions is (identity, created_at), so every epoch-ranged read scans.
+CREATE INDEX idx_commissions_epoch ON commissions (epoch);
```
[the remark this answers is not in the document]
ilmari-h: existing convention in rest of SQL, ignoring

api/src/handlers/list_validators.rs
```diff
@@ -1166,23 +1181,6 @@ mod tests {
...
-        let incidents = Material::default().skipped("skipper", 100).build();
-        let config = GetValidatorsConfig {
-            query_incident_free: Some(true),
-            ..config()
-        };
-        assert_eq!(
-            vote_accounts(filter_validators(validators, &incidents, &config)),
-            vec!["clean".to_string()]
-        );
-    }
```
ochaloup, line 1184: 🤖 claude: This test and the two deleted further down (`a_restart_under_the_floor_does_not_carry_its_epoch_s_block_production`, `a_block_production_incident_outside_the_window_is_dropped_like_any_other`) still pass as written — the `carried` refactor in `into_response_incidents` builds the same set of filtered downtime epochs the old collect did, and `filter_validators`, `config()` and the `Material` helpers are untouched. Removing them drops the only coverage that a sub-floor restart does not suppress its epoch's block production, and that block-production incidents honour `query_incident_free` and the incident window.

store/src/incidents.rs
```diff
@@ -23,12 +23,16 @@ pub const MAX_SKIP_RATE_THRESHOLD: f64 = 0.05;
 /// Under this many seconds a `DOWN` interval is restart noise, not an incident.
 pub const DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS: u64 = 180;
 
+// Validator raised commission to this or more in an epoch -> incident
+pub const COMMISSION_SPIKE_THRESHOLD_PERCENTAGE: u8 = 90;
```
ochaloup, line 27: again, should not be this an argument rather than a const?

store/src/dto.rs
```diff
@@ -477,6 +477,16 @@ pub enum IncidentDetail {
         epoch_end_at: DateTime<Utc>,
         block_production: BlockProductionDetail,
     },
+    /// An epoch the validator raised its inflation commission to 90% or above in
```
ochaloup: I don't like having a "value" in comment - should not be rather a pointer to a const or just be a bit vague on the "value" to what this is above?

ochaloup (changes requested):
Few points, please check them 🙏

api/src/handlers/list_validators.rs
ilmari-h: the paths this was testing are already covered by other tests so thats why this was deleted, did some cleanup. This is a block production incident variant specific test and there are now more general tests for all the variants.

store/src/incidents.rs
ilmari-h: As there are more and more incident types coming, I wanted to keep the API interface from growing out of control with new query params constantly added.
this 90% was agreed on as a good number from data dug by Jarda, so I think we will go with this for a while. If it needs to be changed, then I will expose the argument from the API.

store/src/dto.rs
ilmari-h: agreed, fixed!

ochaloup (approved):
🐊 :rocket:

PR #278, opened 2026-09-15
[GEN-9115] Late client version patch incident type

ilmari-h:
Added incident type `LateClientVersionPatch`

To be merged after https://github.com/marinade-finance/delegation-strategy-2/pull/277

## Summary by CodeRabbit

- **New Features**
  - Added detection and reporting for validators running a late client version.
  - Reports include the client version, lineage, epoch range, newer-version stake share, and per-version stake shares.
  - Added support for retrieving these incidents through validator incident APIs.
  - Incidents are evaluated only for lineages with at least 10 valid members.

- **Documentation**
  - Added the new incident type and version stake-share data to the generated API schema.

- **Tests**
  - Added coverage for detection, lineage-size requirements, stake comparisons, serialization, and API responses.


store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+            let lineage = newer_stake_shares.get(&(stats.epoch, client_lineage.clone()))?;
+            let newer: Vec<dto::VersionStakeShare> = lineage
+                .iter()
+                .take_while(|(above, _)| *above > version)
+                .map(|(above, share)| dto::VersionStakeShare {
+                    version: above.as_str().to_string(),
+                    share: *share,
+                })
+                .collect();
+            let newer_stake_share = newer.iter().map(|above| above.share).sum();
```
ochaloup, line 242: 🤖 claude: Summing per-version f64 shares makes the 80% bar depend on how the newer stake is split — 800 of 1000 lineage stake spread over eight versions sums to 0.7999999999999999 and raises no incident, while the same 800 on a single version does. Keep the comparison in `Decimal` (sum the newer stake, test `newer * 100 >= total * 80`) and convert to f64 only for the response.

store/src/utils.rs
```diff
@@ -381,6 +380,17 @@ pub async fn load_validator_incidents(
         }
     }
 
+    let newer_stake_shares = crate::incidents::newer_stake_shares(records.values());
+    for (vote_account, record) in records {
+        let late_patch = crate::incidents::running_late_client_version_incidents(
+            crate::incidents::epochs_running_late_client_version(record, &newer_stake_shares),
+            from_epoch..=last_epoch,
```
ochaloup, line 385: 🤖 claude: `from_epoch` is `last_epoch + 1 - DEFAULT_CACHE_EPOCHS` and `load_validators` loads exactly those same 90 epochs, so the window's oldest epoch has no predecessor in `breached` and can never open an incident — a validator behind for all 90 is served 89 of them. Either load one epoch below `from_epoch` for the grace lookback, or drop the claim in `epochs_running_late_client_version`'s doc comment that a predecessor outside the window is available.

store/src/incidents.rs
```diff
@@ -244,6 +378,26 @@ impl ValidatorIncidentRecords {
...
+            {
+                incidents.push(dto::IncidentRecord {
+                    epoch: late_patch.epoch,
+                    detail: dto::IncidentDetail::RunningLateClientVersion {
+                        epoch_start_at: late_patch.epoch_start_at,
+                        epoch_end_at: late_patch.epoch_end_at,
+                        version: late_patch.version.clone(),
+                        client_lineage: late_patch.client_lineage.clone(),
+                        newer_stake_share: late_patch.newer_stake_share,
+                        newer_version_stake_shares: late_patch.newer_version_stake_shares.clone(),
```
ochaloup, line 402: 🤖 claude: `filter_validators` builds this for every cached validator before any limit is applied, and `newer_version_stake_shares` is uncapped — with the tens of live agave versions mainnet carries per epoch that is ~20 heap `String`s per record, up to 89 records per validator, across ~1500 validators on each `query_incident_types=RunningLateClientVersion` request. Worth capping the list by share or truncating to the top N, computing `newer_stake_share` over the full prefix first so the reported total stays honest.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+impl EpochClientVersion {
+    pub fn counts_as_incident(&self) -> bool {
+        self.newer_stake_share >= MIN_NEWER_VERSION_STAKE_SHARE
+    }
+}
+
+/// `None` for an epoch still running, a client id absent from the registry, or a version string
+/// nothing can parse.
+fn epoch_client_version(stats: &dto::ValidatorEpochStats) -> Option<(ValidatorVersion, String)> {
+    let version = stats.version.as_deref()?.parse().ok()?;
```
ochaloup, line 171: 🤖 claude: `validators.version` is nullable and a single-epoch gossip gap leaves it NULL, which keeps that epoch out of `breached` and closes the run — so a validator whose version alternates observed/NULL never accumulates two consecutive epochs and is never flagged, however long it stays behind. Worth a bounded carry-forward from the last observation, or at minimum a note that an alternating gap is a blind spot.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+    pub client_lineage: String,
+    /// Share of the lineage's stake on a strictly newer version, as a fraction.
+    pub newer_stake_share: f64,
+    pub newer_version_stake_shares: Vec<dto::VersionStakeShare>,
+}
+
+impl EpochClientVersion {
+    pub fn counts_as_incident(&self) -> bool {
+        self.newer_stake_share >= MIN_NEWER_VERSION_STAKE_SHARE
+    }
```
ochaloup, line 165: 🤖 claude: nit — the bar has no floor on the population behind it, so a two-node lineage like `sig` flags its smaller member automatically once the other upgrades: one peer holding 900 against 100 reads as `newer_stake_share: 0.9`. Consider gating the lineage-epoch on a minimum stake or member count before any share is computed.

store/src/incidents.rs
```diff
@@ -26,13 +28,16 @@ pub const DEFAULT_MIN_INCIDENT_DOWNTIME_SECONDS: u64 = 180;
 // Validator raised commission to this or more in an epoch -> incident
 pub const COMMISSION_SPIKE_THRESHOLD_PERCENTAGE: u8 = 90;
 
+pub const MIN_NEWER_VERSION_STAKE_SHARE: f64 = 0.80;
```
ochaloup, line 31: 🤖 claude: nit — this is a second answer to "which version should this validator be on", alongside the SFDP and feature-gate floors `releases.rs` already publishes at `/releases`. The two can contradict each other: a validator meeting the published agave floor of 4.1.0 still earns an incident once 80% of lineage stake moves past it, so the response doc should say the bar is peer adoption rather than an obligation.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+        self.newer_stake_share >= MIN_NEWER_VERSION_STAKE_SHARE
+    }
+}
+
+/// `None` for an epoch still running, a client id absent from the registry, or a version string
+/// nothing can parse.
+fn epoch_client_version(stats: &dto::ValidatorEpochStats) -> Option<(ValidatorVersion, String)> {
+    let version = stats.version.as_deref()?.parse().ok()?;
+    let lineage = stats.client_lineage.clone()?;
+    stats.epoch_start_at?;
```
ochaloup, line 173: 🤖 claude: nit — this can never fire: `load_validators` defaults `epoch_start_at` to `Utc::now()` before pushing the stats, so it is always `Some`. The `epoch_end_at?` line below is the one doing real work.

```suggestion
```

store/src/utils.rs
```diff
@@ -381,6 +380,17 @@ pub async fn load_validator_incidents(
         }
     }
 
+    let newer_stake_shares = crate::incidents::newer_stake_shares(records.values());
+    for (vote_account, record) in records {
```
ochaloup: 🤖 claude: nit — this is a second full pass over the same ~1500-entry map right after the block-production loop above, and nothing in it depends on that loop. Hoisting `newer_stake_shares` next to `cluster_skip_rates` lets this body fold into the existing loop.

ochaloup (dismissed):
All nice to me. Claude has got several findings, but all are minor ones.

store/src/incidents.rs
ochaloup, line 242: 🤖 claude: Summing the per-version f64 shares loses the bar — a lineage exactly 80% ahead spread over eight versions sums to 0.7999999999999999, so `counts_as_incident` returns false and no incident is ever opened, while the identical 80% sitting on one version does fire. Sum the newer stake in `Decimal` and divide once, converting to f64 only for the response field.

`the_bar_is_read_at_eighty_percent` only exercises a single newer version, so nothing currently covers this.

store/src/utils.rs
ochaloup, line 385: 🤖 claude: The grace rule needs `from_epoch - 1`, but `load_validators` fills `epoch_stats` with the same 90 epochs this range spans, so the window's oldest epoch can never clear the predecessor check and is silently never served — it also shifts `order_field=incidents` and `query_incident_free`. Note `load_commission_raises` reads `from_epoch - 1` for exactly this reason; this path has the same need and no equivalent lookback.

store/src/incidents.rs
ochaloup, line 402: 🤖 claude: `newer_version_stake_shares` holds one heap `String` per distinct newer version — tens per epoch after a release — for up to 89 epochs per validator, and this clone runs for every cached validator even when the caller asked for `limit=1`. Cap the list (drop shares under ~0.1%, or keep the top N by share) and compute `newer_stake_share` over the full prefix before truncating, so the reported total stays honest.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+
+impl EpochClientVersion {
+    pub fn counts_as_incident(&self) -> bool {
+        self.newer_stake_share >= MIN_NEWER_VERSION_STAKE_SHARE
+    }
+}
+
+/// `None` for an epoch still running, a client id absent from the registry, or a version string
+/// nothing can parse.
+fn epoch_client_version(stats: &dto::ValidatorEpochStats) -> Option<(ValidatorVersion, String)> {
```
ochaloup, line 170: 🤖 claude: `version` is nullable per epoch and the collector writes `None` whenever the answering RPC's `getClusterNodes` view misses the node, so a single NULL epoch closes the run — and a validator whose rows alternate observed/NULL never accumulates two consecutive epochs and is never flagged at all, however long it stays behind. Either carry the last observation forward for a bounded number of epochs, or name the blind spot in the doc on `running_late_client_version_incidents`.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+    pub version: String,
+    pub client_lineage: String,
+    /// Share of the lineage's stake on a strictly newer version, as a fraction.
+    pub newer_stake_share: f64,
+    pub newer_version_stake_shares: Vec<dto::VersionStakeShare>,
+}
+
+impl EpochClientVersion {
+    pub fn counts_as_incident(&self) -> bool {
+        self.newer_stake_share >= MIN_NEWER_VERSION_STAKE_SHARE
```
ochaloup, line 164: 🤖 claude: nit — this bar has no floor on the population it is measured against, unlike `MIN_LEADER_SLOTS` and `MIN_MISSED_SLOTS` on the sibling incident types, so a two-node `sig` lineage flags its smaller member off the evidence of one peer. Gate the `(epoch, lineage)` bucket on a minimum lineage stake or validator count before any share is computed.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+    let version = stats.version.as_deref()?.parse().ok()?;
+    let lineage = stats.client_lineage.clone()?;
+    stats.epoch_start_at?;
+    stats.epoch_end_at?;
+    Some((version, lineage))
+}
+
+/// Keyed by epoch and lineage, then by the version the share is measured from. A validator
+/// `epoch_client_version` rejects weighs on neither side.
+pub fn newer_stake_shares<'a>(
```
ochaloup, line 181: 🤖 claude: nit — this is a second, stake-derived answer to "which version should a validator be on", and it will disagree with the published floors `/releases` serves: a validator meeting every floor still earns an incident once 80% of its lineage moves ahead. Cheapest fix is to say in the `RunningLateClientVersion` doc that the bar is peer adoption rather than a required version, so the two endpoints cannot be read as contradicting each other.

store/src/incidents.rs
ochaloup, line 173: 🤖 claude: nit — this guard is dead: `load_validators` defaults `epoch_start_at` to `Utc::now()` before pushing, so it is never `None` on any input that reaches here. The `epoch_end_at` line below does real work — it keeps the in-flight epoch out of the stake denominator — and should stay.

```suggestion
```

store/src/utils.rs
ochaloup: 🤖 claude: nit — this walks the same ~1500-entry `records` map the block-production loop just finished walking, and nothing in it depends on that loop; `newer_stake_shares` is already hoisted above it exactly as `cluster_skip_rates` is hoisted above the first. Fold this body into the earlier loop and compute both maps together.

store/src/incidents.rs
```diff
@@ -138,6 +144,133 @@ pub struct CommissionRaise {
...
+        .filter(EpochClientVersion::counts_as_incident)
+        .collect();
+
+    late_patch.sort_by_key(|epoch| epoch.epoch);
+    late_patch
+}
+
+/// An epoch counts only where the one before it was over the bar too. An epoch with nothing usable
+/// reported is no breach, so it closes the run.
+pub fn running_late_client_version_incidents(
```
ochaloup, line 263: 🤖 claude: nit — this and `epochs_running_late_client_version` are only ever called as one pipeline, and calling the first alone returns every over-the-bar epoch including the ones this grace excludes, with nothing in its name or return type saying the result is unfinished. Merging them also flattens the nested `crate::incidents::` call at the single call site.

store/src/incidents.rs
ilmari-h: very rare occurence and unlikely/almost impossible chance of happening so leaving this

store/src/utils.rs
ilmari-h: fixed the doc comment here

store/src/incidents.rs
ilmari-h: i see the point here but we already have similar heavy per-validator-array load, e.g. /validators response body `epoch_stats`. Yes, this is bounded, but newer_version_stake_shares is also basically bounded: there will rarely be more than a handful of newer client versions than what a given validator is running, unless the validator is seriously out of date. Not so concerned with this so ignoring

store/src/incidents.rs
ilmari-h: store/src/validators.rs:80 already carries the version forward within the epoch row, so NULL needs a whole epoch without gossip.

store/src/incidents.rs
ilmari-h: 🤖 wontfix: the guard keeps `epoch_client_version` one filter for both passes. `newer_stake_shares` and `epochs_running_late_client_version` have to accept the same epoch stats. Without it a row with no `epoch_start_at` adds stake to the lineage total but can never produce an incident. The guard states the contract of the `Option` field, so it does not depend on the `Utc::now()` default in `load_validators`.

store/src/utils.rs

ilmari-h:
@ochaloup thanks, fixed some of the comments. rebased from master so need new approval please

ochaloup (approved):

PR #281, opened 2026-09-17
[GEN-9071] [GEN-9072] collect activating and deactivating stake amounts

ilmari-h:
Added `activating_stake` and `deactivating_stake` to `validators` table, populated by the collector.

Also added new `order_field` option `ActivatingStake` as required by task GEN-9072.

## Summary by CodeRabbit

- **New Features**
  - Validator records now report activating and deactivating stake separately when available.
  - Stake totals include validator self stake, bond amounts, and pending activation or deactivation activity.
  - Validator groups now aggregate and display activating stake.
  - Validators and groups can be sorted by activating stake, with unavailable values placed last.
  - Stake information is persisted consistently with other validator data.

- **Bug Fixes**
  - Missing stake values remain unavailable rather than being replaced with zero.
  - Stake totals safely handle validators without matching stake-account data.


store/src/validators.rs
```diff
@@ -105,7 +105,9 @@ pub async fn store_validators(
             inflation_rewards_commission_bps = CASE WHEN u.inflation_rewards_commission_bps_is_v4 IS NOT NULL THEN u.inflation_rewards_commission_bps ELSE validators.inflation_rewards_commission_bps END,
             inflation_rewards_commission_bps_is_v4 = COALESCE(u.inflation_rewards_commission_bps_is_v4, validators.inflation_rewards_commission_bps_is_v4),
             block_revenue_commission_bps = CASE WHEN u.inflation_rewards_commission_bps_is_v4 IS NOT NULL THEN u.block_revenue_commission_bps ELSE validators.block_revenue_commission_bps END,
-            pending_delegator_rewards = CASE WHEN u.inflation_rewards_commission_bps_is_v4 IS NOT NULL THEN u.pending_delegator_rewards ELSE validators.pending_delegator_rewards END
+            pending_delegator_rewards = CASE WHEN u.inflation_rewards_commission_bps_is_v4 IS NOT NULL THEN u.pending_delegator_rewards ELSE validators.pending_delegator_rewards END,
+            activating_stake = u.activating_stake,
+            deactivating_stake = u.deactivating_stake
```
ochaloup: 🤖 claude: These two are the only optional columns on this UPDATE written unconditionally — `version`, `client_id_raw` and the six v4 columns all guard against an absent value. A snapshot whose `activating_stake` is `None` (the `#[serde(default)]` case) nulls the stored amounts, which drops the validator out of the group sum entirely rather than contributing a zero.

```suggestion
            activating_stake = COALESCE(u.activating_stake, validators.activating_stake),
            deactivating_stake = COALESCE(u.deactivating_stake, validators.deactivating_stake)
```

collect/src/solana_service.rs
```diff
@@ -932,18 +945,25 @@ fn process_accounts_for_self_stake(
...
                     effective,
-                    activating: _,
-                    deactivating: _,
+                    activating,
+                    deactivating,
                 } = stake_account
                     .stake()
                     .unwrap()
                     .delegation
                     .stake_activating_and_deactivating(epoch, stake_history, None);
```
ochaloup, line 954: 🤖 claude: The `None` here is `new_rate_activation_epoch`, so `warmup_cooldown_rate` returns the legacy 0.25 for every epoch while mainnet runs 0.09 — in an epoch where cluster-wide activating stake exceeds the cap, `activating` comes out too low and `effective` too high. That was invisible while both values were discarded; they are now stored, served and sorted on. `marinade_service.rs:110` passes the same `None`.

collect/src/solana_service.rs
```diff
@@ -932,18 +945,25 @@ fn process_accounts_for_self_stake(
...
+                let is_self_stake = withdraw_authorities
+                    .contains(&(withdrawer_key, vote_key.clone()))
+                    && effective != 0;
+                if !is_self_stake && activating == 0 && deactivating == 0 {
+                    continue;
+                }
+                let totals = totals.entry(vote_key).or_default();
+                totals.activating += activating;
+                totals.deactivating += deactivating;
+                if is_self_stake {
```
ochaloup, line 964: 🤖 claude: This new branching has no test, while the sort and the group sum both got one — including the case that forced the assert above to change from `!is_empty()` to `any(|t| t.self_stake != 0)`: a non-self account with `activating > 0` now creates an entry whose `self_stake` stays zero. The file's existing `mod tests` could cover it with a hand-built `StakeStateV2` and a small `StakeHistory`.

collect/src/solana_service.rs
```diff
@@ -963,25 +983,20 @@ fn get_withdrawer_and_vote_keys(stake_account: &StakeStateV2) -> Option<(String,
     })
 }
 
-fn update_self_stake(self_stake: &mut HashMap<String, u64>, vote_key: &str, lamports: u64) {
-    let stake_entry = self_stake.entry(vote_key.to_string()).or_insert(0);
-    *stake_entry += lamports;
-}
-
-pub fn fetch_self_stake(
+pub fn fetch_stake_account_totals(
```
ochaloup: 🤖 claude: nit — the only caller is `get_stake_account_totals` in this same file, so this needn't be exported from the crate.

```suggestion
fn fetch_stake_account_totals(
```

ochaloup (dismissed):
This seems straightforward. Approving.

Consider few findings by Claude.

collect/src/solana_service.rs
ilmari-h: 🤖 The rate is correct, the effect is nil at current mainnet numbers.

The rate only limits when cluster activating stake goes over 9 percent of effective stake, about 35M SOL. Real volume is a few million SOL per epoch, so `newly_effective_cluster_stake / cluster_activating` stays above 1 and each account activates in full in one epoch under either rate.

Left as `None` here and in `marinade_service.rs`. If activation volume ever comes near the cap, the fix is `Some(0)`, since every epoch the collector reads is after the feature activation.

store/src/validators.rs
ilmari-h: fixed

collect/src/solana_service.rs
ilmari-h: fixed

collect/src/solana_service.rs
ilmari-h: fixed

ilmari-h:
@ochaloup thanks, fixed the comments except for 1

store/src/validators.rs
ochaloup: 🤖 claude: Fixed — `COALESCE(u.activating_stake, validators.activating_stake)` keeps the stored amount when a snapshot omits the field, so the validator still contributes to its group sum instead of dropping out.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Declining — your numbers check out: at ~390M SOL effective the 0.09 cap is ~35M SOL per epoch, an order of magnitude above real activation volume, so both rates settle an account in one loop iteration and yield identical amounts.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Fixed — `mod stake_account_tests` covers all four branches, including the non-self account whose entry keeps `self_stake` at zero.

collect/src/solana_service.rs
ochaloup: 🤖 claude: Fixed — `fetch_stake_account_totals` is private now, and clippy is clean so nothing outside the module was relying on the export.

ochaloup (approved):

PR #282, opened 2026-09-17
[GEN-8946][close] finalization of epoch commission adjustment

ochaloup:
`close_epoch` writes `commission_effective` only for validators it finds in the performance snapshot, so a vote account that exists in the epoch's `validators` rows but not in the snapshot keeps a NULL rate — the APY path then reads it as 100% commission. The in-memory "unresolved" tally counts snapshot members only, so it reported zero while such a row sat there.

- Fill `commission_effective` from the sampled vote state for validators the snapshot never listed, so a vote account created between the two collector reads is no longer treated as a 100% commission validator
- Run that fill before `update_observed_commission` so the new rates feed `commission_min_observed` and `commission_max_observed` like every other validator's
- Surface a failed fill only after uptimes and the observed-commission floor have landed, since a closed epoch is never re-listed and those steps would otherwise be skipped
- Count the rows still lacking `commission_effective` in the database and warn, so leftovers outside the snapshot become visible
- Export `close_epoch` from the store lib crate and cover the new write with DB-backed tests, pinning epoch scoping and the precedence of an existing reward-row rate

## Bug Fixes

* Improved epoch closure reliability when validators fall outside the performance snapshot.
* Epoch finalization now continues when sampled commission-rate storage fails for unlisted validators.
* Preserved commission-floor data when an outside-snapshot update cannot be written.
* Ensured reward precedence and epoch scoping remain correct during finalization.
* Added clearer reporting for validator records that remain unresolved.
* Clarified how vote-state commission rates are determined from validator voting samples.


store/src/close_epoch.rs
```diff
@@ -322,8 +365,39 @@ pub async fn close_epoch(
         );
     }
 
+    let (outside_vote_accounts, outside_rates): (Vec<&str>, Vec<i32>) = sampled_commission_bps
+        .iter()
+        .filter(|(vote_account, _)| !snapshot.validators.contains_key(*vote_account))
+        .map(|(vote_account, bps)| (vote_account.as_str(), i32::from(bps_to_percent(*bps))))
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — `snapshot.validators` and `snapshot.rewards` are both built from the same `get_vote_accounts()` result with no filter on either side (`collect/src/validators_performance.rs:219-253` and `:266-279`), so an account cannot be in `rewards` while absent from `validators`. Reward-row precedence is already pinned for the case that can occur, by `the_outside_write_leaves_a_reward_row_and_an_older_epoch_alone`.

store/src/close_epoch.rs
```diff
@@ -60,6 +60,49 @@ async fn load_sampled_commission_bps(
...
+    psql_client: &Client,
+    epoch: &Decimal,
+    vote_accounts: &[&str],
+    rates: &[i32],
+    updated_at: &DateTime<Utc>,
+) -> anyhow::Result<()> {
+    psql_client
+        .execute(
+            "UPDATE validators
+             SET commission_effective = u.commission_effective,
```
ilmari-h, line 74: **🤖 Medium: the outside write overwrites an existing rate**

This UPDATE has no `commission_effective IS NULL` guard, and it stamps `vote_state` as the source unconditionally. The snapshot path uses `resolve_commission_effective`, where a reward row wins. The two paths disagree.

On a re-run of a closed epoch, a row that migration `0029-commission-effective-backfill.sql` filled with `commission_advertised` = 10 becomes `div_ceil(1001, 100)` = 11. `update_observed_commission` then lifts `commission_max_observed` to 11 and crosses the 10 percent line that `load_ruggers` tests in `store/src/utils.rs:515`.

Add `AND validators.commission_effective IS NULL` to the WHERE clause, so the write is a fill only.

store/src/close_epoch.rs
```diff
@@ -322,8 +365,37 @@ pub async fn close_epoch(
         );
     }
 
+    let (outside_vote_accounts, outside_rates): (Vec<&str>, Vec<i32>) = sampled_commission_bps
```
ilmari-h, line 369: **🤖 Low: a row with no sample stays NULL forever**

`sampled_commission_bps` holds only the rows with a non-NULL `inflation_rewards_commission_bps`. A `validators` row outside the snapshot with no sample gets no write here, and the epoch never re-opens. The `UNSAMPLED` test pins that NULL as intended.

Migration `0029` used `commission_advertised` for exactly this case. The new warning shows the row, but no code fixes it. Is `commission_advertised` an acceptable last fallback here?

store/src/close_epoch.rs
```diff
@@ -322,8 +365,37 @@ pub async fn close_epoch(
...
+                "Effective commission from sampled vote state for {} validators the snapshot did not list",
+                outside_vote_accounts.len()
+            ),
+            Err(err) => warn!(
+                "Could not store effective commission for validators outside the snapshot: {err}"
+            ),
+        }
+    }
+
     update_uptimes(psql_client, snapshot.epoch).await?;
```
ilmari-h, line 395: **🤖 Low: the comment above claims more than the code does**

Line 373 says the floor must land even when the outside write fails. This line still uses `?`. `create_epoch_record` already wrote the `epochs` row, so a failure here skips `update_observed_commission`, and `ls-open-epochs` never lists the epoch again. The floor is lost with no recovery path.

The invariant holds for one of the two steps only. Either drop the `?` here too, or narrow the comment.

store/src/close_epoch.rs
ochaloup: 🤖 claude: Fixed — the outside write is now fill-only (`AND validators.commission_effective IS NULL`), so it can never restamp a rate a reward row or the 0029 backfill already resolved, and it stays correct if `create_epoch_record` ever gains `ON CONFLICT (epoch)`. The `load_ruggers` crossing is a separate, pre-existing ceil/floor split between `bps_to_percent` and agave's floored `commission_advertised` — it fires on the snapshot path too, so it is deferred to the commission-projection work.

store/src/close_epoch.rs
ochaloup: 🤖 claude: Declining — `commission_advertised` is the gameable field and 0029 was a one-off backfill rather than the forward contract, so the NULL plus the new warning is the wanted behaviour. `load_ruggers` gates the epoch on `commission_min_observed IS NOT NULL`, not on `commission_effective`, so the epoch still contributes its floor.

store/src/close_epoch.rs
ochaloup: 🤖 claude: Declining — the comment scopes itself to the outside write two lines below it, and that write is swallowed by the `match`, so the floor does land when it fails. The `?` on `update_uptimes` is master's behaviour untouched by this branch; the epoch-level recovery gap it describes is the same unrecoverable-close problem that the `ON CONFLICT (epoch)` decision covers.

store/src/close_epoch.rs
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — `snapshot.rewards` and `snapshot.validators` are built from the same `vote_accounts.current.chain(delinquent)` iterator in `validators_performance.rs`, so an account absent from `validators` has no reward row to prefer. The suggested regression case with differing reward-row and sampled rates cannot be produced by a real snapshot.

ochaloup:
@ilmari-h :pray: re-review

1. Fixed. The outside write is now fill-only (AND validators.commission_effective IS NULL), so a re-close can never restamp a rate that a reward row or the 0029 backfill already resolved.
2. Deferred. commission_effective ceils from bps while commission_advertised floors, so load_ruggers can read a steady 10.01–10.99% rate as a rise past the cap — pre-existing, fires on the snapshot path too, and fixing it means changing the rounding of the whole commission_* family.
3. Declined. Falling back to commission_advertised for an unsampled row would re-introduce the gameable field; the NULL plus the new warning is the wanted behaviour, and load_ruggers gates on commission_min_observed, so the epoch is not dropped.

store/src/close_epoch.rs
```diff
@@ -60,6 +60,50 @@ async fn load_sampled_commission_bps(
...
+               AND validators.commission_effective IS NULL",
+            &[
+                &vote_accounts,
+                &rates,
+                &COMMISSION_EFFECTIVE_SOURCE_VOTE_STATE,
+                epoch,
+                updated_at,
+            ],
+        )
+        .await?;
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — no trigger or CHECK exists on `commission_effective`, `commission_effective_source` or `updated_at`, so one row cannot reject the batch; the only such constraint is the one `a_failed_outside_write_still_leaves_the_epoch_its_floor` adds to its own throwaway schema. Per-row writes would trade an all-or-nothing statement for partial writes on an epoch that is never re-listed, and `warn_on_unresolved_commission` already counts the exact remainder from the database.

ilmari-h (approved):

PR #283, opened 2026-09-18
[GEN-8969][api] adding direct staking collector and api

ochaloup:
**Problem:** `/validators` reports liquid, native and institutional stake per validator, but the ~101k SOL delegated through the direct-staking authority is invisible.
**Core change:** `get_direct_stakes` scans the `psrStL2…` staker authority on each collection and the per-validator total lands in a new `direct_stake` column.

- Collect per-validator stake from the direct-staking authority alongside the existing liquid, native and institutional scans
- Expose it as an additive `direct_stake` field, leaving `marinade_stake` liquid-only so no existing consumer's numbers shift
- Register the authority in `/static/config` as `Marinade Single Validator`, matching `stake_pools`, so a consumer building an authority→product map does not attribute direct stake to an unknown third party
- Log the collected sum and validator count, since an empty scan writes 0 for every validator and would otherwise leave no trace
- Default the snapshot field to 0 on deserialize, so a snapshot written by the previous release still stores across a redeploy
- Cover the INSERT and UPDATE store paths with a round-trip test, since the new column renumbered the UPDATE type-index map

## Summary by CodeRabbit

* **New Features**
  * Added direct-stake tracking for validators, including current totals and epoch-level statistics.
  * Added Marinade Single Validator delegation authority details to the static configuration response.
  * Direct-stake values are now retained when validator records are created or updated.

* **Documentation**
  * Updated validator API examples to include direct-stake values for current records and epoch statistics.

* **Tests**
  * Added coverage confirming direct-stake values are stored, loaded, and preserved alongside existing stake data.


migrations/0032-direct-stake.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE validators ADD COLUMN direct_stake NUMERIC NOT NULL DEFAULT 0;
```
ilmari-h, line 1: Is defaulting to 0 correct or would it be more true to reality to have a NULL value here?

E.g.  I recently added `activating_stake` and `deactivating_stake` and decided to make those NULLable, so historical values that the collector doesn't write to are not falsely showing "0".

ilmari-h (changes requested):
Just 1 question about the SQL nullability, looks good.

migrations/0032-direct-stake.sql
ochaloup: Hm... ok, nullable seems more correct here.

ochaloup: 🤖 claude: Fixed — `direct_stake` is now a nullable column (no default), carried as `Option` through collector, store and API, so epochs before collection return `null` instead of `0`. The UPDATE uses `COALESCE`, like `activating_stake`, so a snapshot without the field keeps the stored value.

ochaloup:
@ilmari-h updated :pray: re-review

collect/src/validators.rs
```diff
@@ -324,6 +332,7 @@ pub fn collect_validators_info(
             institutional_stake: *institutional_stake.get(&vote_pubkey).unwrap_or(&0),
             activating_stake: Some(stake_totals.activating),
             deactivating_stake: Some(stake_totals.deactivating),
+            direct_stake: Some(*direct_stake.get(&vote_pubkey).unwrap_or(&0)),
```
ilmari-h, line 335: **🤖 An empty scan writes 0 for every validator, and the 0 stays**

This line maps a missing key to `Some(0)`, so the collector never sends a NULL. The UPDATE guard `COALESCE(u.direct_stake, validators.direct_stake)` only keeps a NULL, so one bad run overwrites the stored stake with 0 for all validators.

The authority holds 16 stake accounts and 101,656 SOL today, so an empty scan means a rotated authority, not a real zero. `get_foundation_stakes` asserts a non-empty map for this reason.

Fix: assert a non-empty map, or send `None` when the map is empty.

ilmari-h (commented):
found just 1 thing that I think is worth checking

collect/src/validators.rs
ochaloup: 🤖 claude: Fixed with the `None` option — an empty direct scan now sends `None` for every validator and logs a warning, so `COALESCE` keeps the stored value and a new epoch reads NULL (unknown) instead of 0. I left out the assert because testnet has no direct authority and a fully exited authority is valid, so it would crash the collector on correct data.

ochaloup:
@ilmari-h can you re-review please? I updated the PR a bit :sweat: (sorryyy... :-) as I've got some points from validators hub what is good to have for help to show the direct staking data better

PR #285, opened 2026-09-18
[GEN-9062] filtering `/validators` by client or provider name

ilmari-h:
## Summary by CodeRabbit

- **New Features**
  - Added provider and client filters to validator queries.
  - Filters are case-insensitive and can match unknown or unregistered providers and clients.
  - Client filtering recognizes both client lineage and block-engine labels.
  - Blank filter values are ignored.
  - Provider, client, and operator grouping now consistently handles unknown values under an unclassified category.


store/src/groups.rs
```diff
@@ -117,6 +124,27 @@ fn folded(key: &Option<String>) -> FoldedKey {
...
+}
+
+/// Whether `validator` belongs to the `/clients` row named `key`, or to one of the block engine
+/// rows under it. Case-insensitive.
+pub fn belongs_to_client(validator: &ValidatorRecord, key: &str) -> bool {
+    let key = folded(&normalized(Some(key.to_string())));
+
+    [GroupKind::ClientLineage, GroupKind::ClientLabel]
+        .into_iter()
+        .any(|kind| folded(&current_client_key(validator, kind)) == key)
```
ochaloup: 🤖 claude: Client ids 0 and 3 spell both the lineage and the label `Agave`, so `.any()` over the two levels makes the bare `Agave` child row that `/clients` publishes unselectable — `query_client=Agave` returns the whole lineage (Jito, Paladin, JitoBAM, Rakurai, Raiku) against a row that counted only plain Agave. Same collision for `Frankendancer`, `Firedancer` and the `Unknown` child; matching `ClientLineage` only here and adding a separate `query_block_engine` for the label level would keep every `/clients` row addressable.

api/src/handlers/list_validators.rs
```diff
@@ -87,6 +87,10 @@ pub struct QueryParams {
     query_verified: Option<bool>,
     query_protected: Option<bool>,
     query_flagged: Option<bool>,
+    /// Query by a provider name, as returned by `/providers`
+    query_provider: Option<String>,
+    /// Query by a client name, as returned by `/clients`
+    query_client: Option<String>,
```
ochaloup: 🤖 claude: These doc comments are the whole OpenAPI contract for the new params, and they omit that the match is exact and case-insensitive and that `Unknown` selects the no-value bucket. `/providers?query=` is a substring match, so a consumer who narrows with `query=hetz` and reuses `hetz` here gets zero rows and no error.

```suggestion
    /// Exact, case-insensitive match on a `key` from `/providers`. `Unknown` selects the validators with no recorded provider.
    query_provider: Option<String>,
    /// Exact, case-insensitive match on a `key` from `/clients`. `Unknown` selects the validators with no recorded client.
    query_client: Option<String>,
```

api/src/handlers/list_validators.rs
```diff
@@ -959,6 +982,133 @@ mod tests {
...
+            vec!["jito".to_string()]
+        );
+    }
+
+    #[test]
+    fn query_client_unknown_keeps_the_validators_of_an_unregistered_client() {
+        let validators = map(vec![
+            running_client("agave", "agave", "Agave"),
+            validator("unregistered", 100, vec![]),
+        ]);
```
ochaloup, line 1187: 🤖 claude: This fixture has `client_id_raw: None`, so it is a validator with no client info at all rather than an unregistered one, and the `client_id_raw` fallback in `current_client_key` that `belongs_to_client` depends on stays untested — deleting that `or_else` leaves every test in this PR green. A second assertion that `query_client=Sonic` also keeps it would close the branch.

```suggestion
        let validators = map(vec![
            running_client("agave", "agave", "Agave"),
            ValidatorRecord {
                client_id_raw: Some("Sonic".to_string()),
                ..validator("unregistered", 100, vec![])
            },
        ]);
```

api/src/handlers/list_validators.rs
```diff
@@ -959,6 +982,133 @@ mod tests {
         );
     }
 
+    fn hosted_by(vote_account: &str, aso: &str) -> ValidatorRecord {
+        let mut validator = validator(vote_account, 100, vec![]);
+        for stats in validator.epoch_stats.iter_mut() {
+            stats.dc_aso = Some(aso.to_string());
+        }
+        validator
+    }
```
ochaloup, line 1007: 🤖 claude: Every fixture gets the same `dc_aso` on both epochs, so the provider tests pass whether `belongs_to_provider` pins `epoch` or scans any epoch — dropping the epoch filter would leave them green while `/validators` silently stops agreeing with `/providers`. A fixture carrying `Hetzner Online GmbH` at epoch 99 and `TeraSwitch` at epoch 100 is the case that catches it.

ochaloup (approved):
It seems ok to me. Claude has got minor findings. Just check, in case I can re-approve instantly when you decide to fix some of them.

store/src/groups.rs
ilmari-h: yes this is a great point, adding...

api/src/handlers/list_validators.rs
ilmari-h: done

api/src/handlers/list_validators.rs
ilmari-h: not valid after recent commit

api/src/handlers/list_validators.rs
```diff
@@ -87,6 +90,12 @@ pub struct QueryParams {
     query_verified: Option<bool>,
     query_protected: Option<bool>,
     query_flagged: Option<bool>,
+    /// Exact, case-insensitive match on a `key` from `/providers`. `Unknown` selects the validators with no recorded provider.
+    query_provider: Option<String>,
+    /// Exact, case-insensitive match on a `key` from `/clients`. `Unknown` selects the validators with no recorded client.
+    query_client: Option<String>,
+    /// Exact, case-insensitive match on a `key` from a `/clients` child row, one client paired with one block engine, such as `Agave + Jito`.
```
ochaloup: 🤖 claude: nit — `/clients` publishes an `Unknown` child row under the `Unknown` parent, so `query_block_engine=Unknown` works, but this doc is the whole OpenAPI contract and it is the one of the three that does not say so.

```suggestion
    /// Exact, case-insensitive match on a `key` from a `/clients` child row, one client paired with one block engine, such as `Agave + Jito`. `Unknown` selects the validators with no recorded block engine.
```

store/src/groups.rs
```diff
@@ -117,6 +124,24 @@ fn folded(key: &Option<String>) -> FoldedKey {
     key.as_ref().map(|key| key.to_lowercase())
 }
 
+/// Whether `validator` belongs to the `/providers` row named `key`. Case-insensitive.
```
ochaloup: 🤖 claude: This reads the record-level `dc_aso` while `/providers` keys off `stats.dc_aso` at `population.current_epoch`, and the two agree only because `filter_validators` runs `is_eligible_validator` first — a second caller that skips that gate silently gets a different population. The `hosted_by` fixture sets `dc_aso` on the record but leaves every epoch stat `None`, a state `load_validators` cannot produce, so no test pins the agreement either.

```suggestion
/// Whether `validator` belongs to the `/providers` row named `key`, which needs `is_eligible_validator` to have passed. Case-insensitive.
```

ochaloup (dismissed):

api/src/handlers/list_validators.rs
ilmari-h: this works differently now. You query for no-block-engine with 'none'

store/src/groups.rs
ilmari-h: improved the comment

store/src/groups.rs
```diff
@@ -117,6 +124,32 @@ fn folded(key: &Option<String>) -> FoldedKey {
...
+
+/// The `key` that selects the validators running no block engine.
+pub const NO_BLOCK_ENGINE: &str = "none";
+
+/// Whether `validator` runs the block engine whose vendor slug is `key`. `NO_BLOCK_ENGINE` selects
+/// the validators running none. Case-insensitive.
+pub fn belongs_to_block_engine(validator: &ValidatorRecord, key: &str) -> bool {
+    let key = key.trim().to_lowercase();
+    if client_engine(validator.client_id).is_none() {
+        return key == NO_BLOCK_ENGINE;
```
ochaloup, line 147: 🤖 claude: `client_engine` is `None` for an unregistered client too, so `none` returns every unknown client (e.g. raw `Sonic`) as running no block engine, and the `Sonic` / `Unknown` child rows under the `/clients` `Unknown` parent can no longer be selected on their own. Limiting `none` to registered clients keeps it honest; the `none` test would then expect only `plain`.

```suggestion
    if client_engine(validator.client_id).is_none() {
        return key == NO_BLOCK_ENGINE && validator.client_lineage.is_some();
```

collect/src/solana_service.rs
```diff
@@ -176,42 +201,55 @@ impl ClientId {
...
+    }
+
+    pub fn engine(&self) -> Option<&'static str> {
+        self.groupings().and_then(|grouping| grouping.engine)
     }
 
-    pub fn label(&self) -> Option<&'static str> {
-        self.groupings().map(|(_, _, label)| label)
+    /// The lineage as display text, and the block engine after it: `Agave + Jito`.
+    pub fn label(&self) -> Option<String> {
```
ochaloup, line 216: 🤖 claude: nit — `label()` now builds a `String`, and `classified()` in `store/src/dto.rs` calls it only to test `is_some()`, so every `client_*` helper allocates per validator per epoch in `load_validators`. Gating `classified` on `client.lineage().is_some()` gives the same answer without the allocation.

ochaloup (dismissed):

collect/src/solana_service.rs
ilmari-h: seems to just save some small allocation and potentially makes this less readable so skipping

store/src/groups.rs

ochaloup (approved):

